PR #1, opened 2026-02-20
feat: Remove all JS related code from this codebase

AbhiPrasad:
This PR removes all js related code from this codebase, as it was a remanent of the transition from https://github.com/braintrustdata/braintrust-sdk.

ref https://github.com/braintrustdata/braintrust/pull/11071

plan: https://www.notion.so/braintrustdata/Splitting-up-braintrust-sdk-30df7858028980298731d3827a23cb70


AbhiPrasad:
CI is skipped here because we only run it when python files change https://github.com/braintrustdata/braintrust-sdk-python/blob/ec0c35e2fadc7bd6a734ff3013226193b1f81d09/.github/workflows/py.yaml#L5-L11

clutchski (approved):

PR #14, opened 2026-02-26
Evals(py): Allow passing tags for evals and experiments

Mahhheshh:
as said in: https://github.com/braintrustdata/braintrust-sdk/pull/1397#issuecomment-3960510051

;p
savemypr


AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.7.0

PR #23, opened 2026-03-03
fix: Make ADK and pydantic integration wrappers thread safe

ibolmo:


py/src/braintrust/wrappers/test_pydantic_ai_integration.py
```diff
@@ -2639,3 +2703,67 @@ class TestAutoInstrumentPydanticAI:
     def test_auto_instrument_pydantic_ai(self):
         """Test auto_instrument patches Pydantic AI and creates spans."""
         verify_autoinstrument_script("test_auto_pydantic_ai.py")
+
+
+def test_model_request_stream_sync_thread_context_propagation(memory_logger, direct):
```
AbhiPrasad: I think we can mark this with `@pytest.mark.vcr`

AbhiPrasad (approved):

Issue #22, opened 2026-03-02
Add pytest integration


AbhiPrasad:
Similar to what we have with vitest in the JavaScript SDK: https://www.braintrust.dev/docs/integrations/sdk-integrations/vitest

If you want to see this supported, please upvote this issue and leave a comment with your use case. Thanks!

PR #25, opened 2026-03-03
feat: Add pytest integration

AbhiPrasad:
resolves https://github.com/braintrustdata/braintrust-sdk-python/issues/22

## AI Summary

This adds a pytest plugin to the Braintrust SDK that automatically tracks test results as Braintrust experiments.

### How it works

Mark tests with `@pytest.mark.braintrust` and run with the `--braintrust` flag:

```python
import pytest

@pytest.mark.braintrust
def test_my_llm(braintrust_span):
    result = my_llm("hello")
    braintrust_span.log(input={"query": "hello"}, output=result, scores={"relevance": 0.95})
    assert "greeting" in result
```

```bash
pytest --braintrust --braintrust-project="my-project"
```

Each marked test becomes a span in a Braintrust experiment. Pass/fail is automatically scored (1.0/0.0). At the end of the session, the plugin prints the standard SDK experiment summary and flushes results:

```
=========================SUMMARY=========================
95.00% 'pass'  score

See results for my-project at https://www.braintrust.dev/app/...
```

Without `--braintrust`, the `braintrust_span` fixture returns a no-op span — tests run normally with zero overhead and no API calls.

### Design choices

- **Activation**: Requires `--braintrust` flag. This is safer — no surprise API calls when you just have braintrust installed and run `pytest`. This follows the pattern of pytest-django which requires explicit configuration before it does anything.
- **Logging API**: The `braintrust_span` fixture exposes the native `Span` object directly — no wrapper. Users call `span.log(input=..., output=..., scores=...)`, the same API used everywhere else in the SDK. This avoids teaching a plugin-specific API and lets users batch fields into a single `log()` call to minimize merge overhead. Users can also access `braintrust.current_span()` from deeper in the call stack when needed.
- **Auto-input logging**: Test function arguments from `@pytest.mark.parametrize` are auto-logged as input.
- **Config**: CLI options (`--braintrust-project`, etc.) with env var fallback. CLI options are more discoverable via `pytest --help`.
- **Summary formatting**: Reuses the SDK's existing `ExperimentSummary.__str__()` rather than a custom box-drawing format, keeping output consistent with `braintrust.init()` / the eval framework.
- **Entry point registration**: The plugin auto-loads via `pytest11` entry point when braintrust is installed — no `conftest.py` configuration needed. pytest is not declared as a dependency since the entry point is inert unless pytest is already installed (standard pattern used by celery, django, and flask).

### Pytest version support

We chose **pytest >= 8** as the minimum supported version. The plugin uses `pytester` subprocess mode and stable pytest APIs (`hookimpl`, `pytest_configure`, `pytest_runtest_setup/teardown/makereport`, `@pytest.fixture`, `@pytest.mark`). We test against pytest 8.4.2 and latest in CI.

### Key behaviors

**Static data via marker** — `input`, `expected`, `tags`, and `metadata` can be passed as marker kwargs instead of logging them in the test body:

```python
@pytest.mark.braintrust(
    input={"query": "hello"},
    expected={"answer": "world"},
    tags=["regression"],
    metadata={"model": "gpt-4"},
)
def test_with_metadata(braintrust_span):
    result = my_llm("hello")
    braintrust_span.log(output=result)
```

**Auto-input from parametrize** — `@pytest.mark.parametrize` args are automatically logged as input. Each parametrized case becomes its own span:

```python
@pytest.mark.braintrust
@pytest.mark.parametrize("query,expected_answer", [
    ("2+2?", "4"),
    ("Capital of France?", "Paris"),
])
def test_qa(braintrust_span, query, expected_answer):
    result = my_llm(query)
    braintrust_span.log(output=result)
    # Auto-logged as input: {"query": "2+2?", "expected_answer": "4"}
```

If both `@pytest.mark.parametrize` and marker `input` are provided, marker `input` takes precedence.

**Experiment grouping** — one experiment is created per test module by default. Override globally with `--braintrust-project` or per-test/class with the marker:

```python
@pytest.mark.braintrust(project="my-project")
class TestMyLLM:
    def test_case_a(self, braintrust_span): ...
    def test_case_b(self, braintrust_span): ...
```

**Failure tracking** — failed tests automatically log the error string to the span alongside the `pass: 0.0` score:

```
FAILED test_example.py::test_bad - AssertionError: expected "hello"
```
```python
# Span automatically receives:
#   scores: {"pass": 0.0}
#   error: "AssertionError: expected \"hello\"\n..."
```

**Summary control** — suppress the terminal summary with `--braintrust-no-summary`:

```bash
pytest --braintrust --braintrust-no-summary
```

### DIFF
--- a/py/noxfile.py
+++ b/py/noxfile.py
@@ -75,7 +75,7 @@ DSPY_VERSIONS = (LATEST,)
 GOOGLE_ADK_VERSIONS = (LATEST, "1.14.1")
 # temporalio 1.19.0+ requires Python >= 3.10; skip Python 3.9 entirely
 TEMPORAL_VERSIONS = (LATEST, "1.20.0", "1.19.0")
-PYTEST_VERSIONS = (LATEST, "8.4.2", "7.4.4")
+PYTEST_VERSIONS = (LATEST, "8.4.2")
 
 
 @nox.session()
--- a/py/src/braintrust/wrappers/pytest_plugin/README.md
+++ b/py/src/braintrust/wrappers/pytest_plugin/README.md
@@ -5,7 +5,7 @@ Automatically track pytest test results as [Braintrust](https://www.braintrust.d
 ## Requirements
 
 - Python >= 3.10
-- pytest >= 7
+- pytest >= 8
 
 ## Installation
 


py/src/braintrust/wrappers/pytest_plugin/plugin.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Exposed to tests via the ``braintrust_span`` fixture.
+    """
+
+    _span: Span
+
+    def __init__(self, span: Span) -> None:
+        self._span = span
+
+    def log_input(self, input: Any) -> None:
```
clutchski: i understand the reasoning for this wrapping, but i'm a bit skeptical (a) it teaches users an API that they can't use everywhere (b) it triggers more work. spans must be merged on the client side (kinda expensive) or on the server side (more expensive) because you can't possibly log input/output/expected in one call.

PR #40, opened 2026-03-04
chore: Exclude test packages from setuptools

colinbennettbrain:
Fix braintrust-langchain shipping tests/ as an installed top-level package, which breaks pytest discovery in downstream projects.


AbhiPrasad (approved):
Thanks!

PR #43, opened 2026-03-05
enhance BraintrustSpanProcessor to support custom exporters

colinbennettbrain:
This change makes `BraintrustSpanProcessor` support custom OpenTelemetry exporters in addition to the default Braintrust export path.  Some users need to route traces through their existing OTEL pipeline (e.g. vendor-specific exporters, internal collectors, or multi-destination export) while still using Braintrust instrumentation. Previously, `BraintrustSpanProcessor` did not make that customization path explicit/easy.

- Added `exporter=None` to:
  - `BraintrustSpanProcessor(...)`
  - `add_braintrust_span_processor(...)`
- If `exporter` is provided, the processor uses it directly.
- In that case, `api_key`, `parent`, `api_url`, and `headers` are ignored.
- If `exporter` is not provided, behavior is unchanged (the default Braintrust `OtelExporter` is still created).

You can create your exporter as usual, then pass it via the new `exporter` argument.

```
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
from braintrust.otel import BraintrustSpanProcessor

custom_exporter = OTLPSpanExporter(endpoint="http://localhost:4318/v1/traces")
processor = BraintrustSpanProcessor(exporter=custom_exporter)
```
Or with the helper:

```
from braintrust.otel import add_braintrust_span_processor
add_braintrust_span_processor(
    tracer_provider=provider,
    exporter=custom_exporter,
)
```

### DIFF
--- a/py/src/braintrust/otel/__init__.py
+++ b/py/src/braintrust/otel/__init__.py
@@ -210,6 +210,7 @@ def add_braintrust_span_processor(
     filter_ai_spans: bool = False,
     custom_filter=None,
     headers: dict[str, str] | None = None,
+    exporter=None,
 ):
     processor = BraintrustSpanProcessor(
         api_key=api_key,
@@ -218,6 +219,7 @@ def add_braintrust_span_processor(
         filter_ai_spans=filter_ai_spans,
         custom_filter=custom_filter,
         headers=headers,
+        exporter=exporter,
     )
     tracer_provider.add_span_processor(processor)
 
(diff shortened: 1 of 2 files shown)


AbhiPrasad (commented):
Can we add more details to the PR description about why this change was needed, and what users have to now do to add custom exporters? This will be useful for the changelogs.

AbhiPrasad:
closing for repo clean up. Please open a new PR if you still want this to get merged in!

PR #61, opened 2026-03-09
Let python SDK handle classifiers

CLowbrow:
Example in the PR

### DIFF
--- a/py/examples/evals/intent_classification_eval.py
+++ b/py/examples/evals/intent_classification_eval.py
@@ -1,0 +1,45 @@
+from braintrust import Eval
+from openai import OpenAI
+
+client = OpenAI()
+
+DATASET = [
+    {
+        "input": "What's your return policy?",
+        "expected": "policy_question",
+    },
+    {
+        "input": "I need help with my order",
+        "expected": "support_request",
+    },
+]
+
+
+def task(input):
+    response = client.responses.create(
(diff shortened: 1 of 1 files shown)


py/src/braintrust/framework.py
```diff
@@ -1358,16 +1415,40 @@ async def await_or_run_scorer(root_span, scorer, name, **kwargs):
...
+                classification_items = _try_parse_classification_output(result)
+                span.log(output=classification_items, classifications={name: classification_items})
+                return {"scores": [], "classifications": {name: classification_items}}
 
             if isinstance(result, Iterable):
-                for s in result:
-                    if not is_score(s):
+                result = list(result)
+                if all(is_score(s) for s in result):
+                    result = list(result)
```
AbhiPrasad: This is redundant with the call above.

py/src/braintrust/framework.py
```diff
@@ -232,7 +232,64 @@ class EvalScorerArgs(SerializableDataClass, Generic[Input, Output]):
     metadata: Metadata | None = None
 
 
-OneOrMoreScores = Union[float, int, bool, None, Score, list[Score]]
+OneOrMoreScores = Union[
```
AbhiPrasad: We should probbaly rename this, it now includes classifiers. Or represent the type in another way.

Issue #67, opened 2026-03-10
Make the Claude Agent SDK support instrumenting hooks


AbhiPrasad:
https://platform.claude.com/docs/en/agent-sdk/hooks

PR #71, opened 2026-03-11
feat: Add support for claude agent sdk hooks

AbhiPrasad:
resolves https://github.com/braintrustdata/braintrust-sdk-python/issues/67

image

This PR adds support for instrumenting claude agent sdk hooks. These are created as separate spans, as they are useful to query and visualize on the trace waterfall alongside tool calls and agent spans.

To make the claude agent sdk hooks instrumentation more robust, this PR also adds extra wrapping around top level module query calls. This should ensure spans are created for more use cases that leverage the claude agent sdk.


viadezo1er:
I think this triggers at UI bug. Tested it in the wrong org, I'll share my trace soon.

PR #72, opened 2026-03-12
ci: Run python publishing via gha trigger

AbhiPrasad:
This PR removes the ability to publish via a script (`py/scripts/push-release-tag.sh`) in favour of entirely driving this via gha triggers. This was based on work I did in the JS repo here: https://github.com/braintrustdata/braintrust-sdk-javascript/pull/1544

This means you can just release via the github UI. `docs/publishing.md` has all the details about how this works.

I also aligned the test pypi and real pypi release process, and made running dry runs a bit more robust.


py/scripts/validate-release.py
AbhiPrasad: I converted this to be a python file. I think we should stick with writing python scripts instead of bash for the sdk repo, easier to use everywhere (and technically makes this more windows compatible for contributors)

PR #74, opened 2026-03-12
ci: Add nightly canary release for python sdk

AbhiPrasad:
Similar to the strategy used here: https://github.com/braintrustdata/braintrust-sdk-javascript/pull/1547, this PR adds a github workflow that publishes a nightly canary release for the braintrust package to Test PyPi. Read through `docs/publishing.md`, it has all the info you need.

We also don't publish github releases or git tags for canaries for now, but we can change that decision at a later date.

We have to do all the publishing from a single workflow because pypi trusted publishing relies on that.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -163,7 +163,7 @@ class EvalHooks(abc.ABC, Generic[Output]):
 
     @property
     @abc.abstractmethod
-    def metadata(self) -> Metadata:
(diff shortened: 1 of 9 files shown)


viadezo1er:
`--run-number n` is manually set by the user, so it's a bit error prone, we might decrement or skip version number.
However I'm not sure the added check required (Github and Pypi and maybe a local registry etc...) outweighs the value of preventing this, especially since we should know what version we want to type.

PR #79, opened 2026-03-13
fix: CachedSpanFetcher no longer caches empty results permanently

cjgalione:
## Summary
- `CachedSpanFetcher.get_spans()` was setting `_all_fetched = True` even when the fetch returned no spans, permanently caching the empty result and preventing future re-fetches
- Now only marks as fully fetched when the cache is non-empty, so subsequent calls will retry and pick up spans once they become available

## Test plan
- Updated `test_handle_empty_results` to verify re-fetch occurs when first call returned empty
- Added `test_empty_then_populated_refetches` to confirm spans are returned on retry
- Added `test_empty_results_with_type_filter` to confirm type-filtered path also handles empty results
- Verified existing caching tests still pass (non-empty fetches are still cached)


AbhiPrasad (dismissed):
If there are a bunch of scorers running, this will mean that `get_spans` will keep hitting the network, but I guess that is fine.

CLowbrow (approved):

CLowbrow (approved):
seems fine but @AbhiPrasad should approve as well.

AbhiPrasad (approved):

AbhiPrasad:
I pushed up some changes to get this in a good state. Will merge in!

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.15.0

PR #85, opened 2026-03-16
chore: Update formatting and linting

viadezo1er:
Update formatters and linters
Remove unused isort, flake8, black config files and import/dependencies
Remove .prettierignore (JS config in python repo)
Add unused isort config to ruff check config
Add ruff format step in pre-commit

The first commit has the config file changes the second commit is just a ```make fixup```.
So to review the PR I highly recommend looking at the first commit, and only afterward briefly look at the second commit.

~~Pylint role is unchanged. Maybe it should also be updated? It would make ```make (py)lint``` would be much faster.~~
Pylint can't be replaced by ruff as of now.


viadezo1er:
Ah forgot to update the CI

PR #102, opened 2026-03-20
fix: Solve dependency issues

viadezo1er:
Fix https://github.com/braintrustdata/braintrust-sdk-python/issues/112

langchain python pkg got updated, some submodules are now accessible with langchain_classic module

See comparison of submodules between [langchain_classic](https://reference.langchain.com/python/langchain-classic/langchain_classic) and [langchain](https://reference.langchain.com/python/langchain/langchain)

setup.py is never used in the dev workflow so that might cause issues since the dependencies used at packaging time are never tested.
The doc extra dependency can probably be deleted, especially since it can cause dependency issues.

### DIFF
--- a/py/src/braintrust/cli/install/redshift.py
+++ b/py/src/braintrust/cli/install/redshift.py
@@ -4,7 +4,7 @@ import re
 from hashlib import md5
 
 # pylint: disable=no-name-in-module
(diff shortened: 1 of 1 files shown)


viadezo1er:
i checked what this says

PR #119, opened 2026-03-23
fix(devserver): add x-bt-use-gateway to CORS allowed headers

cjgalione:
## Summary

- Add `x-bt-use-gateway` to `ALLOWED_HEADERS` in the Python devserver CORS config, matching what `api-ts/src/cors.ts` already allows
- Add a test that verifies CORS preflight accepts this header

## Context

The Braintrust Playground sends `x-bt-use-gateway: true` when gateway routing is enabled. The main data plane (`api-ts`) added this header to its CORS allowlist in Feb (commit `ee5f4f65e0`), but the Python SDK devserver was never updated to match.

This causes browsers to block preflight `OPTIONS` requests to any remote eval server built on `braintrust.devserver`, with an error like:

```
Request header field x-bt-use-gateway is not allowed by Access-Control-Allow-Headers
```

Multiple demo projects (`pydantic-supervisor`, `langgraph-supervisor`, `google-adk-supervisor`) independently worked around this by monkey-patching `bt_cors.ALLOWED_HEADERS` at startup. This fix addresses it at the source.

## Test plan

- [x] New test `test_cors_preflight_allows_gateway_header` sends an OPTIONS request with `x-bt-use-gateway` and asserts it appears in the response's `access-control-allow-headers`
- [x] All existing devserver integration tests continue to pass

### DIFF
--- a/py/src/braintrust/devserver/cors.py
+++ b/py/src/braintrust/devserver/cors.py
@@ -23,6 +23,7 @@ ALLOWED_HEADERS = [
     "x-bt-project-id",
     "x-bt-stream-fmt",
     "x-bt-use-cache",
+    "x-bt-use-gateway",
     "x-stainless-os",
     "x-stainless-lang",
     "x-stainless-package-version",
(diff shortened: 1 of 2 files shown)


AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.11.0

Issue #116, opened 2026-03-23
Adopt common integrations interface for the integrations the Python SDK supports


AbhiPrasad:
Instead of wrappers with different kinds of apis, let's re-organize this repo to use integrations.

This should make it easier to understand what we support, and how patching works. Inspired from https://github.com/braintrustdata/braintrust-sdk-ruby/tree/main/lib/braintrust/contrib

PR #132, opened 2026-03-24
ref(agno): Make agno use new integrations API

AbhiPrasad:
ref https://github.com/braintrustdata/braintrust-sdk-python/issues/116

Convert the Agno instrumentation from the legacy wrappers pattern to the new integrations API introduced in https://github.com/braintrustdata/braintrust-sdk-python/pull/118.

I also heavily edited the agent skills with some of the learnings I made while doing this. Should become much more effective now.


viadezo1er (approved):
stamp

PR #142, opened 2026-03-25
migrate braintrust langchain into main pkg

viadezo1er:
# DRAFT

 Migration Guide: ```braintrust-langchain``` to ```braintrust```

### Before (old)
```bash
pip install braintrust-langchain
```

```python
from braintrust_langchain import BraintrustCallbackHandler, set_global_handler
handler = BraintrustCallbackHandler()
set_global_handler(handler)
```

### After (new)
```bash
pip install braintrust
```

#### Option 1: Auto-instrument (recommended)
```python
from braintrust import auto_instrument
auto_instrument()
```

### What stays the same

- All public APIs (BraintrustCallbackHandler, BraintrustTracer, set_global_handler, clear_global_handler) work identically
- Existing from braintrust_langchain import ... imports still work but emit a DeprecationWarning
- Full support for chains, LLMs, tools, retrievers, agents, LangGraph, streaming, and async

### What to update

- Change imports from braintrust_langchain to braintrust.integrations.langchain (or use auto_instrument())
- Remove braintrust-langchain from your dependencies — braintrust now includes LangChain support with no extra package

### DIFF
--- a/integrations/langchain-py/src/tests/test_callbacks.py
+++ b/integrations/langchain-py/src/tests/test_callbacks.py
@@ -340,9 +340,7 @@ def test_tool_usage(logger_memory_logger: LoggerMemoryLogger):
                                 "message": {
                                     "content": ANY,  # May be empty for tool calls
                                     "type": "ai",
-                                    "additional_kwargs": {
-                                        "tool_calls": ANY,  # Tool call details
-                                    },
+                                    "additional_kwargs": ANY,
                                     "response_metadata": ANY,
                                 },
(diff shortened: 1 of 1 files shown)


py/src/braintrust/wrappers/langchain/__init__.py
AbhiPrasad: We don't need this file. We don't need to include wrappers for backwards compatibility. Users can just import from `braintrust.integrations.langchain` for everything.

Issue #145, opened 2026-03-25
Add instrumentation for AgentScope


AbhiPrasad:
https://github.com/agentscope-ai/agentscope

AbhiPrasad:
agentscope has otel support, but we can document that as well

PR #147, opened 2026-03-25
feat: Add agentscope integration for tracing

AbhiPrasad:
resolves https://github.com/braintrustdata/braintrust-sdk-python/issues/145

image


py/src/braintrust/integrations/agentscope/patchers.py
viadezo1er: Factory function instead of 10 almost identical classes?

PR #167, opened 2026-03-27
Fix Google GenAI input serialization for Content objects and tool fallback

colinbennettbrain:
This fixes contents serialization for types.  Content objects by recursively serializing nested parts, so binary Part.inline_data inside Content.parts is preserved as Braintrust attachments instead of falling through generic serialization. It also hardens tool serialization by keeping Google’s serializer as the primary path, but adding debug  logging and a best-effort fallback when that path fails, so tool definitions are still  logged in structured form rather than being silently dropped.

The change includes regression tests covering nested binary content serialization and tool fallback behavior.


AbhiPrasad:
Were you not able to use VCR tests? I can help generate them if needed, let me know!

AbhiPrasad:
I opened https://github.com/braintrustdata/braintrust-sdk-python/pull/276 with this change refactored a bit + VCR tests. Closing this PR as a result.

AbhiPrasad:

PR #194, opened 2026-04-01
fix(invoke): encode request body as UTF-8 bytes to prevent Latin-1 corruption

evanmkeith:
## Summary

  `invoke()` passed the JSON body as a Python `str` to `requests.post(data=...)`,
  which encodes `str` bodies as Latin-1 by default. Any non-Latin-1 character in
  the payload (em dashes, smart quotes, etc.) either raised a `UnicodeEncodeError`
  or was silently corrupted before the request was sent.

  - Encode `bt_dumps()` output to UTF-8 bytes before passing to `data=`
  - Add `Content-Type: application/json` header (was missing; only `Accept` was set)

  `bt_dumps` is kept — it handles Pydantic models, dataclasses, and NaN/Inf values
  that stdlib `json` cannot serialize. Other SDK paths (`logger.py`) already use
  `.encode("utf-8")` correctly; this brings `invoke()` in line.

  ## Test plan

  - [ ] Unit test: assert `data=` arg to `requests.post` is `bytes`, contains correct UTF-8 encoding of em dash (`\xe2\x80\x94`), and `Content-Type` header is set
  - [ ] Manually run `invoke()` with Unicode input (`"result \u2014 excellent"`) and confirm no `UnicodeEncodeError` and payload reaches the API intact

  Fixes BT-4620

### DIFF
--- a/py/src/braintrust/functions/cassettes/test_invoke_encodes_body_as_utf8_bytes.yaml
+++ b/py/src/braintrust/functions/cassettes/test_invoke_encodes_body_as_utf8_bytes.yaml
@@ -1,0 +1,29 @@
+interactions:
+- request:
+    body: '{"api_version": 1, "input": {"text": "result \u2014 excellent"}, "metadata":
+      null, "parent": "", "project_name": "test-project", "slug": "test-fn", "stream":
+      false, "tags": null}'
+    headers:
+      Accept:
+      - application/json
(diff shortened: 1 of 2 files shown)


AbhiPrasad (commented):
can we add a VCR test instead of mocking if possibe? If it's too annoying no worries.

PR #227, opened 2026-04-08
include sandbox runtime in list request

nselvidge:
In order to add support for python based sandboxes we need to pass the runtime into the list request. The change is backwards compatible and the value will be stripped on out of date data plane versions.


AbhiPrasad (approved):

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.14.0!

PR #262, opened 2026-04-13
feat(litellm: instrument rerank

Mahhheshh:
Instruments `litellm.rerank` api;

ref (#165)

Tests:
Running the test will raise an Invalid API key error as I didn't have a key, so I manually modified the cassette's fields. This is is not acceptable for the final version; if I can get my hands on real HTTP traffic for this API, I will update the PR.

Summary:
image

### DIFF
--- a/py/src/braintrust/integrations/litellm/patchers.py
+++ b/py/src/braintrust/integrations/litellm/patchers.py
@@ -7,10 +7,12 @@ from braintrust.integrations.base import FunctionWrapperPatcher
 from .tracing import (
     _acompletion_wrapper_async,
     _aembedding_wrapper_async,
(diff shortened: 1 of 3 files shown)


AbhiPrasad:
I think we prob wanna add cohere support before we do this (https://github.com/braintrustdata/braintrust-sdk-python/issues/260).

cohere does offer free evaluation keys if you want to give that a try! https://github.com/braintrustdata/braintrust-sdk-python/blob/main/.agents/skills/sdk-integrations/SKILL.md should be the exact guide to follow to add a new integration. Otherwise I'll add the new integration in a bit and then we can look at this PR again. Going to close for now!

PR #274, opened 2026-04-14
feat: add per-input trial_count support to Eval()

AbhiPrasad:
Allow each EvalCase to specify its own trial_count that overrides the global evaluator-level trial_count. Resolution order: datum.trial_count > evaluator.trial_count > 1 (default).

python implementation of https://github.com/braintrustdata/braintrust-sdk-javascript/pull/1814


lforst (approved):

lforst:
ty for being attentive

PR #283, opened 2026-04-14
ci(checks): measure session weights on windows with python 3.10

AbhiPrasad:
Switch the update-session-weights workflow to run on windows-2025 with Python 3.10 instead of ubuntu-latest with Python 3.13. Windows tests are the bigger CI bottleneck, so shard balancing should optimise for their durations. Using the oldest supported Python version targets the slowest likely combination.

Also pin the update job runner to ubuntu-24.04 for consistency with checks.yaml.

### DIFF
--- a/.github/workflows/update-session-weights.yaml
+++ b/.github/workflows/update-session-weights.yaml
@@ -12,8 +12,8 @@ permissions:
 
 jobs:
   measure:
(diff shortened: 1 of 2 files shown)


viadezo1er (approved):
stamp

PR #285, opened 2026-04-14
fix: use Mapping for Metadata and ParametersSchema type aliases

willfrey:
## Summary

- Changes `Metadata` (`types/__init__.py`) and `ParametersSchema` (`parameters.py`) from `dict[str, Any]` to `Mapping[str, Any]`.
- Both aliases are used exclusively in parameter positions across the SDK (`Eval()`, `EvalAsync()`, `init()`, `Logger.log()`, `Dataset.insert()`, `invoke()`, `validate_json_schema()`, etc.). No call site mutates values received through these types.
- Fixes dict invariance: a caller with `dict[str, str]` (valid metadata) was rejected by type checkers because `dict[str, str]` is not assignable to `dict[str, Any]` due to invariance. `Mapping` is covariant in its value type, so this works correctly.
- Follows the same pattern as PR #281 (`EvalParameters` → `Mapping`).

`Score.metadata` is typed as `Metadata` on a dataclass field — verified that no code mutates it after construction.

## Test plan

- [ ] Existing tests pass (annotation-only change, no runtime impact)
- [ ] `dict` subtype callers (e.g. `dict[str, str]`) are now accepted by type checkers

### DIFF
--- a/py/src/braintrust/parameters.py
+++ b/py/src/braintrust/parameters.py
@@ -38,7 +38,7 @@ JSONValue = None | bool | int | float | str | list["JSONValue"] | dict[str, "JSO
 ValidatedParameters = dict[str, object]
 ParameterSchema = PromptParameter | ModelParameter | type[object] | None
 EvalParameters = Mapping[str, ParameterSchema]
-ParametersSchema = dict[str, Any]
+ParametersSchema = Mapping[str, Any]
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


AbhiPrasad (approved):
I regressed this actually, it was Mapping before. Just forgot about the invariance stuff ://

willfrey:
The built-in concrete types are a trap for annotations (usually) ;)

AbhiPrasad:

PR #289, opened 2026-04-14
fix: use Mapping for dict params, Any for ValidatedParameters

willfrey:
## Summary

Three type annotation fixes in `framework2.py` and `parameters.py`:

- **`choice_scores: dict[str, float]` → `Mapping[str, float]`** in `ScorerBuilder.create()` (all 3 overloads). Input parameter — dict invariance means `dict[str, int]` was rejected even though `int` is compatible with `float`.
- **`metadata: dict[str, Any]` → `Mapping[str, Any]`** in `CodeParameters` dataclass field and `ParametersBuilder.create()`. Input parameter — same invariance class as the `Metadata` alias change in PR #285. Other metadata params in this file already use the `Metadata` alias.
- **`ValidatedParameters = dict[str, object]` → `dict[str, Any]`**. Values are Pydantic models, strings, `Prompt` objects, etc. Using `object` forces every caller to narrow; `Any` is the practical type for this heterogeneous container.

## Test plan

- [ ] Existing tests pass (annotation-only changes, no runtime impact)
- [ ] `dict[str, int]` is now accepted for `choice_scores` by type checkers
- [ ] Retrieved `ValidatedParameters` values no longer require narrowing from `object`

### DIFF
--- a/py/src/braintrust/framework2.py
+++ b/py/src/braintrust/framework2.py
@@ -1,6 +1,6 @@
 import dataclasses
 import json
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Mapping, Sequence
 from typing import Any, overload
 
 import slugify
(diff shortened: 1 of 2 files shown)


AbhiPrasad (approved):

AbhiPrasad:

PR #296, opened 2026-04-15
test(context): replace pytest-forked with subprocess-based isolation

AbhiPrasad:
pytest-forked was never installed in nox test sessions (missing from BASE_TEST_DEPS), so the forked mark was unknown on every platform, producing PytestUnknownMarkWarning spam and silently losing the process isolation the context propagation tests rely on. The package is also deprecated and uses os.fork() which is unavailable on Windows.

Replace with subprocess.run-based isolation: each isolated test now spawns a fresh `python -c` subprocess that sets up fixtures, optionally calls setup_threads(), and runs the scenario function. This works cross-platform, provides true process isolation, and has no new dependencies.

Remove pytest-forked from requirements-dev.txt.

### DIFF
--- a/py/requirements-dev.txt
+++ b/py/requirements-dev.txt
@@ -7,7 +7,6 @@ pylint==4.0.5
 pyperf==2.10.0
 pytest==9.0.2
 pytest-asyncio==1.3.0
-pytest-forked==1.6.0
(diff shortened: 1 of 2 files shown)


viadezo1er (approved):
stamp

PR #298, opened 2026-04-15
ref(integrations): migrate test_utils and cassettes out of wrappers/

AbhiPrasad:
Delete 16 stale cassettes from wrappers/cassettes/ left behind by prior google_genai and pydantic_ai migrations. Move the 3 active test_auto_* cassettes (anthropic, google_genai, openrouter) into their respective integration cassette directories.

Move autoinstrument_test_context, verify_autoinstrument_script, run_in_subprocess, and assert_metrics_are_valid from braintrust.wrappers.test_utils into braintrust.integrations.test_utils, then delete the old module. Update all auto_test_scripts and integration test files to import from the new location.

Each auto_test_script now explicitly passes cassettes_dir pointing at its integration's cassettes/ directory instead of relying on the now-removed global CASSETTES_DIR / BRAINTRUST_CASSETTES_DIR fallback.


viadezo1er (approved):
stamp

Issue #256, opened 2026-04-10
switch to using uv + pyproject.toml instead of setup.py + pin all dependencies


AbhiPrasad:
Should vastly improve repo dx and supply chain security in the repo. Every single dep needs to be pinned.

We need to figure out how to still test against latest integration package versions in the noxfile test suite.

PR #299, opened 2026-04-15
test(integrations): per-version VCR cassette directories

AbhiPrasad:
In prep for https://github.com/braintrustdata/braintrust-sdk-python/issues/256, we record vcr cassettes per version. Large diff, but that's just because we now have these duplicate cassettes. Also re-org noxfile for these changes.


viadezo1er (approved):
stamp

PR #300, opened 2026-04-15
build(nox): centralize dependency pins in pyproject.toml

AbhiPrasad:
Migrate all nox session dependency pins and package metadata into py/pyproject.toml as the single source of truth:

- Replace setup.py with pyproject.toml (setuptools backend)
- Move base test deps, session auxiliary deps, lint deps, build deps, and dev deps into PEP 735 dependency groups
- Move provider version matrices into [tool.braintrust.matrix] custom table (read by noxfile at import time)
- Generate and commit uv.lock for reproducible auxiliary dep resolution
- Pin all previously-floating 'latest' versions to explicit versions
- Remove requirements-dev.txt, requirements-build.txt, requirements-optional.txt, and the install-optional Makefile target
- Switch Makefile and CI from uv pip install to uv sync
- Bump uv to 0.11.6 in .tool-versions
- Add daily dependency-updates.yml workflow that runs uv lock --upgrade and labels PRs based on whether provider SDKs changed
- Exclude uv.lock from codespell checks (false positive on astroid)
- Update AGENTS.md, CONTRIBUTING.md, and agent skill docs to reflect the new layout


viadezo1er (approved):
stamp

PR #306, opened 2026-04-16
Widen BraintrustStreamChunk and annotate BraintrustStream.stream

### LINKED ISSUE #294: BraintrustStream.stream untyped; _parse_sse_stream yields chunks outside declared union
## Problem

Two entangled type issues in `functions/stream.py`:

### 1. `self.stream` has no type annotation

`BraintrustStream.__init__` assigns `self.stream` without an annotation (`stream.py:99-101`). Type checkers cannot reason about the attribute type.

### 2. `_parse_sse_stream` return type is too narrow

`_parse_sse_stream` declares return type `Generator[BraintrustStreamChunk, None, None]`, but the function body yields `BraintrustConsoleChunk` and `BraintrustProgressChunk` in addition to the three types in the union:

```python
BraintrustStreamChunk = Union[BraintrustTextChunk, BraintrustJsonChunk, BraintrustErrorChunk]
```

The generator actually yields 5 chunk types, not 3.

## Why these are entangled

Annotating `self.stream` as `Generator[BraintrustStreamChunk, ...] | list[BraintrustStreamChunk]` (the obvious fix for issue 1) would carry forward the return type lie from issue 2. Fixing one without the other makes things worse.

## Options

1. **Widen `BraintrustStreamChunk`** to include all 5 chunk types — simplest, but changes the public type that callers pattern-match against
2. **Create a separate internal union** (e.g. `_InternalStreamChunk`) for the generator, keep `BraintrustStreamChunk` as the public API type — more precise, but adds complexity
3. **Filter console/progress chunks inside `_parse_sse_stream`** and don't yield them — but `parse_stream` explicitly handles them, so they're expected in the stream

Would appreciate guidance on the preferred direction.

willfrey:
## Summary

Fixes #294.

Two entangled type fixes in `functions/stream.py`:

1. **Widen `BraintrustStreamChunk`** — `_parse_sse_stream` yields `BraintrustConsoleChunk` and `BraintrustProgressChunk` in addition to the three types in the union. The union now includes all five chunk types so the declared type matches actual stream output.

2. **Annotate `self.stream`** — previously untyped. Annotated as `Iterable[BraintrustStreamChunk]` to cover the generator from `_parse_sse_stream`, `list` inputs, and iterators produced by `tee` in `copy()`.

Per [maintainer guidance](https://github.com/braintrustdata/braintrust-sdk-python/issues/294#issuecomment-4248464526), widening the public type is preferred over an internal union — the types should reflect the actual API.

## Test plan

- [x] Verify `_parse_sse_stream` yield types are all members of the widened `BraintrustStreamChunk`
- [x] Verify `parse_stream` handles all five chunk types (lines 191-199)
- [x] Verify `self.stream` annotation is compatible with all assignment sites (`__init__`, `copy`)

### DIFF
--- a/py/src/braintrust/functions/stream.py
+++ b/py/src/braintrust/functions/stream.py
@@ -7,7 +7,7 @@ with utility methods to make them easy to log and convert into various formats.
 
 import dataclasses
 import json
-from collections.abc import Generator
+from collections.abc import Generator, Iterable
 from itertools import tee
 from typing import Literal, Union
 
@@ -79,7 +79,13 @@ class BraintrustInvokeError(ValueError):
     pass
 
 
-BraintrustStreamChunk = Union[BraintrustTextChunk, BraintrustJsonChunk, BraintrustErrorChunk]
+BraintrustStreamChunk = Union[
+    BraintrustTextChunk,
+    BraintrustJsonChunk,
+    BraintrustErrorChunk,
+    BraintrustConsoleChunk,
+    BraintrustProgressChunk,
+]
 
 
 class BraintrustStream:
@@ -96,7 +102,7 @@ class BraintrustStream:
             base_stream: Either an SSEClient or a list of BraintrustStreamChunks.
         """
         if isinstance(base_stream, SSEClient):
-            self.stream = self._parse_sse_stream(base_stream)
+            self.stream: Iterable[BraintrustStreamChunk] = self._parse_sse_stream(base_stream)
         else:
             self.stream = base_stream
         self._memoized_final_value = None


willfrey:
One note on the `Iterable[BraintrustStreamChunk]` annotation for `self.stream`:

`self.stream` holds three different runtime types depending on the code path:
1. `Generator[BraintrustStreamChunk, None, None]` — from `_parse_sse_stream` (single-pass)
2. `list[BraintrustStreamChunk]` — from the `else` branch in `__init__` (multi-pass)
3. An iterator from `tee()` — after `copy()` reassigns it (single-pass)

`Iterable` is the only type that's correct for all three, but it doesn't guarantee re-iterability — which is honest, since generators and `tee` iterators *are* single-pass. If someone calls `__iter__()` and then `final_value()` on a generator-backed stream, the second traversal silently yields nothing. That's a pre-existing issue (unrelated to this PR), but worth noting for anyone who wants to tighten the contract later.

AbhiPrasad (approved):

PR #312, opened 2026-04-16
fix(pydantic): retype wrapper spans as task to stop cost double-counting

colinbennettbrain:
Agent wrapper spans (agent_run, agent_run_sync, agent_run_stream*, agent_to_cli_sync, model_request*) were tagged type=llm and logged the same usage metrics as their nested leaf `chat <model>` span. Experiment aggregations that sum metrics across type=llm spans therefore counted a single provider call twice (wrapper + leaf), inflating reported tokens and cost ~2x for a single-turn agent and more for multi-turn runs.

Retag every wrapper span as SpanTypeAttribute.TASK; only the leaf `chat <model>` emitted by _wrap_concrete_model_class stays LLM. _DirectStreamWrapper is used both as wrapper (direct.model_request_stream) and as leaf (Model.request_stream), so it gains a span_type parameter defaulting to LLM; wrapper callers pass TASK explicitly.

Test coverage: flip existing wrapper-type assertions to TASK (leaf chat_span assertions stay LLM) and extend the cassette-backed test_agent_run_async with a regression check that exactly one type=llm span exists and that prompt/completion tokens summed across llm spans equal the leaf's values.

### DIFF
--- a/py/src/braintrust/integrations/pydantic_ai/test_pydantic_ai_integration.py
+++ b/py/src/braintrust/integrations/pydantic_ai/test_pydantic_ai_integration.py
@@ -125,7 +125,7 @@ async def test_agent_run_async(memory_logger):
     assert chat_span is not None, "chat span not found"
 
     # Check agent span
-    assert agent_span["span_attributes"]["type"] == SpanTypeAttribute.LLM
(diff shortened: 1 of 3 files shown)


AbhiPrasad (approved):

AbhiPrasad:

PR #367, opened 2026-04-29
feat(temporal): add temporal integrations auto instrumentation

AbhiPrasad:
Temporal is now available as a first-class integration via braintrust.auto_instrument(), which patches Temporal client and worker setup to install BraintrustPlugin automatically when temporalio is installed.

New applications can enable Temporal tracing with:

```python
import braintrust

braintrust.auto_instrument()
```

For explicit setup without global auto-instrumentation, users can call setup_temporal():

```python
from braintrust.integrations.temporal import setup_temporal

setup_temporal(project_name="my-project")
```

If an application manually constructs Temporal plugin lists, use BraintrustPlugin directly:

```python
from braintrust.integrations.temporal import BraintrustPlugin

plugins = [BraintrustPlugin(), *existing_plugins]
```

Migration guide: replace old contrib imports with integrations imports for new code:

```python
# Before
from braintrust.contrib.temporal import BraintrustPlugin, BraintrustInterceptor

# After
from braintrust.integrations.temporal import BraintrustPlugin, BraintrustInterceptor
```

Prefer braintrust.auto_instrument() for new applications. The braintrust.contrib.temporal module remains as a deprecated compatibility re-export for its previous public API: BraintrustPlugin and BraintrustInterceptor.


lforst (approved):

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.19.0

PR #383, opened 2026-05-01
Adding version_pretty to fetch prompt response

stretpjc:
Adding a `version_pretty` property to loaded prompts so callers can access the prettified transaction/version ID directly from the SDK. This helps since the UI exposes the pretty version only instead of the decimal value that is `_xact_id` by default.

- Added `Prompt.version_pretty`
- Uses the existing `prettify_xact()` helper
- Returns the prettified version when  `_xact_id` is present or `None` if the prompt has no version
- Added a test for this

Tested locally:
image

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -5071,6 +5071,11 @@ class Prompt:
         return self._lazy_metadata.get()._xact_id
 
     @property
+    def version_pretty(self) -> str | None:
+        version = cast(str | None, self.version)
+        return prettify_xact(version) if version is not None else None
+
+    @property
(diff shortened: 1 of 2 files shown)


AbhiPrasad:
I’m undecided if we want to extend the public api this way. For now users can just use `prettify_xact` directly right?

stretpjc:
Yes, that is correct @AbhiPrasad. This would just add it directly to the SDK response so that it is easier for users to cross reference between the UI and SDK

AbhiPrasad:
closing for repo clean up. Please re-open if this PR still applies!

PR #395, opened 2026-05-06
support anthropic ttl token cache metrics

realark:
- add btx test runner for anthropic
- support ttl token metrics for anthropic instrumentation

blocked on this: https://github.com/braintrustdata/braintrust-spec/pull/20


AbhiPrasad:
I've done this in another PR. Should be good to go!

PR #399, opened 2026-05-06
Python parameters typing

j13huang:
Noticed that the typing was not working properly for the parameters docs snippet (https://www.braintrust.dev/docs/evaluate/write-parameters)

I think the main reason is that the `PromptData` type (https://github.com/braintrustdata/braintrust-sdk-python/blob/0f0c1cea8031c80c7a1c1a0aa3eeb7fd36db441c/py/src/braintrust/prompt.py#L68) is not a `TypedDict`. So I set up a separate TypedDict type and codexed my way through supporting that


py/src/braintrust/parameters.py
```diff
@@ -17,12 +18,17 @@
...
+    options: NotRequired[PromptOptionsNullish | None]
+
+
 class PromptParameter(TypedDict):
     """A prompt parameter specification."""
 
     type: Literal["prompt"]
     name: NotRequired[str | None]
-    default: NotRequired[PromptData | None]
+    default: NotRequired[PromptDataDict | None]
```
AbhiPrasad: I think this is a breaking change, we might need to do:
```suggestion
    default: NotRequired[PromptDataDict | PromptData | None]
```

j13huang: Ah yes, thanks

AbhiPrasad (approved):

AbhiPrasad:

PR #403, opened 2026-05-07
fix: Support autoevals inline score typing

j13huang:
I noticed that using autoevals as scorers inline fails typing. Codex says this is because the autoevals scorers return a similar-but-different Score class https://github.com/braintrustdata/autoevals/blob/main/py/autoevals/score.py#L10


AbhiPrasad (approved):
thanks @j13huang!

AbhiPrasad:

PR #423, opened 2026-05-12
fix: update GitPython constraint for security fix

AbhiPrasad:
## Summary
- Require GitPython >= 3.1.50 to pick up the security fix for Dependabot alert 118
- Refresh uv.lock for the patched GitPython release
- Also includes the lockfile-only langchain-core patch release update needed for Dependabot alert 117


realark (approved):

AbhiPrasad:

PR #437, opened 2026-05-15
task(logger): accept serializable objects as metadata

### LINKED ISSUE #435: Allow Span.log metadata to serialize Pydantic models
### Description
Currently when calling `Span.log` the `metadata` property must be a dictionary. The `input` property does not have this restriction and instead goes through `_to_bt_safe` ([here](https://github.com/braintrustdata/braintrust-sdk-python/blob/78d0f2242b8614fdec000667c5f7d11f4fb40599/py/src/braintrust/bt_json.py#L78-L91)) which dumps the model.

It would be nice if the same flow was applied to `metadata`.

### Reproduction
Run the following script:
```
from pydantic import BaseModel
import braintrust
class M(BaseModel):
    foo: str = "bar"
logger = braintrust.init_logger(project="repro", set_current=False)
with logger.start_span(name="x") as span:
    span.log(input=M(), metadata=M())   # input ok, metadata raises
```

#### Expected
It should accept the Pydantic model in the metadata property.

#### Observed
It throws `ValueError: metadata must be a dictionary`.

barrettpyke:
### Description
Run `metadata` through the `bt_safe_deep_copy` before validating it so Pydantic models and other objects that dump to dictionaries can be logged as metadata.
This change covers:
- `Logger.log`
- `Experiment.log`
- `Span.log`
- `Logger.log_feedback`
- `Experiment.log_feedback`
- `update_span`
- `Dataset.__init__`
- `Dataset.insert`
- `Dataset.update`

### Testing
- Added type tests
- Added unit tests for the following:
  - Mocked Pydantic `model_dump` behavior to avoid needing dependency:
    - `span.log`
    - `logger.log`
    - `experiment.log`
    - `logger.log_feedback`
    - `experiment.log_feedback`
    - `dataset.insert`
    - `dataset.update`
  - reject `metadata` with non string keys
  - reject `metadata` that does not serialize into dict
  - test `span.log` with actual pydantic model
- Manually tested the repro script

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -3891,7 +3891,7 @@ class Experiment(ObjectFetcher[ExperimentEvent], Exportable):
         :param expected: (Optional) the ground truth value (an arbitrary, JSON serializable object) that you'd compare to `output` to determine if your `output` value is correct or not.
         :param tags: (Optional) a list of strings that you can use to filter and group records later.
         :param comment: (Optional) an optional comment string to log about the event.
-        :param metadata: (Optional) a dictionary, or an object that serializes to a dictionary, with additional data about the feedback. If you have a `user_id`, you can log it here and access it in the Braintrust UI. Note, this metadata does not correspond to the main event itself, but rather the audit log attached to the event. The values in `metadata` can be any JSON-serializable type, but its keys must be strings.
+        :param metadata: (Optional) a dictionary, or an object that serializes to a dictionary (such as a Pydantic model), with additional data about the feedback. If you have a `user_id`, you can log it here and access it in the Braintrust UI. Note, this metadata does not correspond to the main event itself, but rather the audit log attached to the event. The values in `metadata` can be any JSON-serializable type, but its keys must be strings.
         :param source: (Optional) the source of the feedback. Must be one of "external" (default), "app", or "api".
         """
         return _log_feedback_impl(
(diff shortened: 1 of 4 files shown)


py/src/braintrust/logger.py
```diff
@@ -3381,7 +3391,7 @@ def _log_feedback_impl(
     expected: Any | None = None,
     tags: Sequence[str] | None = None,
     comment: str | None = None,
-    metadata: Metadata | None = None,
```
AbhiPrasad, line 3384: we should update `Metadata` type instead of changing it like this here. Let's also make sure we add type tests for this accordingly.

py/src/braintrust/logger.py
```diff
@@ -2777,7 +2777,22 @@ def _helper(v: Any) -> Any:
...
+
+    for key in metadata.keys():
+        if not isinstance(key, str):
+            raise ValueError("metadata keys must be strings")
+
+    return metadata
+
+
 def _validate_and_sanitize_experiment_log_partial_args(event: Mapping[str, Any]) -> dict[str, Any]:
+    event = dict(event)
```
AbhiPrasad: I would prefer not to shadow the variable here

py/src/braintrust/logger.py
```diff
@@ -2777,7 +2777,22 @@ def _helper(v: Any) -> Any:
     return event
 
 
+def _validate_and_sanitize_metadata(metadata: object) -> dict[str, Any]:
+    if not isinstance(metadata, dict):
+        metadata = bt_safe_deep_copy(metadata)
+
+    if not isinstance(metadata, dict):
```
AbhiPrasad: repeating `if not isinstance(metadata, dict):` feels like we can clean up this logic.

py/src/braintrust/logger.py
barrettpyke: I didn't change it because its used in the Dataset class as well, is that good with you to change the behavior there in addition to the logging methods?

AbhiPrasad: yeah that is fine, the types become more correct

py/src/braintrust/types/__init__.py
```diff
@@ -1,5 +1,13 @@
...
-from typing import Any
+from typing import Any, Protocol, TypeAlias
 
 
-Metadata = Mapping[str, Any]
+class _PydanticV2Metadata(Protocol):
+    def model_dump(self, *, exclude_none: bool = ...) -> Mapping[str, Any]: ...
+
+
+class _PydanticV1Metadata(Protocol):
```
AbhiPrasad: these should prob not be underscore, given they'll be part of the public API w/ `Metadata` now.

AbhiPrasad (approved):

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.22.1

PR #453, opened 2026-05-21
fix(logger): defensively collect no git metadata when org returns nothing

alexr17:
## Summary

When the control plane returns no git_metadata configuration, the SDK now collects nothing instead of substituting SDK-side defaults. This eliminates duplicate default maintenance between SDK and control plane and aligns with feedback from the corresponding JS SDK PR.

## Changes

- Removed `DEFAULT_GIT_METADATA_FIELDS` and `default_git_metadata_settings()` from `git_fields.py`
- Updated `logger.py` `_check_org_info()` to fall back to `collect="none"` when org has no git_metadata
- Updated `gitutil.py` `get_repo_info()` to fall back to `collect="none"` when settings is None
- Updated docstrings in `framework.py` to reflect that collection defers to org-level settings
- Added tests to verify new behavior

## Test plan

- [x] All 3 new unit tests pass
- [x] All 448 existing tests in `test_core` pass
- [x] No regressions


AbhiPrasad (commented):
is this a breaking change for self hosted customers? If it is, is that okay?

alexr17:
> is this a breaking change for self hosted customers? If it is, is that okay?

I'm not sure how breaking it would be but this is the direction we want to go in based on:
https://github.com/braintrustdata/braintrust-sdk-javascript/pull/1972#discussion_r3221999323

AbhiPrasad (approved):

PR #455, opened 2026-05-22
fix(langchain): preserve anthropic cache metrics

AbhiPrasad:
LangChain Anthropic responses report cache reads and cache writes separately from normal input tokens, including TTL-specific cache creation buckets. The previous cached-token fix avoided OpenAI double counting, but it could drop Anthropic cache-write detail from spans and produce totals that were less useful for cost analysis.

Preserve the cache creation metrics users need to understand prompt-cache spend and keep token totals aligned with the prompt-cache semantics, while continuing to avoid double counting OpenAI cached input tokens.

ref https://github.com/braintrustdata/braintrust-spec/blob/main/docs/features/prompt-cache.md

### DIFF
--- a/py/src/braintrust/integrations/langchain/callbacks.py
+++ b/py/src/braintrust/integrations/langchain/callbacks.py
@@ -661,36 +661,40 @@ def _get_metrics_from_response(response: LLMResult):
             input_token_details = usage_metadata.get("input_token_details")
             if input_token_details and isinstance(input_token_details, dict):
                 cache_read = input_token_details.get("cache_read")
-                # langchain-anthropic >= 1.4.0 maps cache_creation_input_tokens to
-                # ephemeral tier fields (ephemeral_5m_input_tokens, ephemeral_1h_input_tokens)
-                # rather than the top-level cache_creation field. Sum both for compat.
                 cache_creation = input_token_details.get("cache_creation")
-                if not cache_creation and (
-                    "ephemeral_5m_input_tokens" in input_token_details
-                    or "ephemeral_1h_input_tokens" in input_token_details
-                ):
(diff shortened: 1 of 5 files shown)


py/src/braintrust/integrations/langchain/callbacks.py
```diff
@@ -661,36 +661,40 @@ def _get_metrics_from_response(response: LLMResult):
...
+                cache_creation_split = (cache_creation_5m or 0) + (cache_creation_1h or 0)
+                if cache_creation is not None or cache_creation_split:
+                    metrics["prompt_cache_creation_tokens"] = max(cache_creation or 0, cache_creation_split)
+                if has_cache_creation_breakdown:
+                    # Anthropic exposes TTL-specific cache creation buckets. Preserve the
+                    # split so downstream cost tooling can price 5m vs 1h writes correctly.
+                    if cache_creation_5m is not None:
+                        metrics["prompt_cache_creation_5m_tokens"] = cache_creation_5m
+                    if cache_creation_1h is not None:
+                        metrics["prompt_cache_creation_1h_tokens"] = cache_creation_1h
```
realark, line 677: If we set the specific TTL tokens, we _don't_ want to set the generic `prompt_cache_creation_tokens`. So I think this block should be if/else instead of two independent if statements

https://github.com/braintrustdata/braintrust-spec/blob/main/test/llm_span/anthropic/prompt_caching_1h.yaml#L259

realark: (to clarify, I mean the outer block that checks `has_cache_creation_breakdown`, not the inner ifs)

realark (approved):

py/src/braintrust/integrations/langchain/callbacks.py
AbhiPrasad: yes for sure! good catch

PR #474, opened 2026-05-28
fix: Parse Retry-After header for 429s

AbhiPrasad:
Previously we actually had two levels of retries for "failed" requests to braintrust:
- The background worker thread would retry if it failed to serialize OR if sending traces to `/logs3` failed
- The http transport layer would retry if we hit http error

This PR cleans that up to remove the background worker thread retries, and make the http transport layer (via the retry adapter) the only source of truth for retries. After that change, it adds support for retrying on 429s to the retry adapter, leveraging the `Retry-After` header to determine when to properly do backoffs.

dependent on https://github.com/braintrustdata/braintrust/pull/14886 deploying as that fixes how we send `Retry-After`

### DIFF
--- a/py/src/braintrust/env.py
+++ b/py/src/braintrust/env.py
@@ -30,6 +30,14 @@ def parse_int(value: str) -> int | None:
         return None
 
 
+def parse_positive_int(value: str) -> int | None:
+    """Parse a non-negative integer from a string."""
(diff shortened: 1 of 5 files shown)


realark (approved):

AbhiPrasad:
gonna close this, I think this is too heavy handed.

PR #486, opened 2026-06-02
fix: honor org git metadata settings for experiments

AbhiPrasad:
Experiments now capture commit and branch metadata when an organization has enabled git metadata collection, even if callers do not pass git_metadata_settings explicitly.

This restores the expected default behavior for orgs collecting all git fields or selected fields, while keeping metadata disabled for orgs without a saved collection policy.


alexr17:
lgtm, i jsut checked that js sdk already has this behavior

lforst (approved):

PR #495, opened 2026-06-04
Keep duplicate devserver evaluators addressable

### LINKED ISSUE #366: `braintrust eval --dev` silently drops evaluators with duplicate `eval_name`
### Summary

When `braintrust eval --dev` loads multiple `eval_*.py` files whose `Eval(...)` calls share a `name` argument, only one evaluator survives the load. The startup log reports the full count loaded, but `GET /list` returns just one entry. There is no warning or error.

### Repro

Two minimal entrypoints, both targeting the same Braintrust project (the natural pattern when you want one project containing multiple evaluators):

`eval_a.py`:
```python
from braintrust import Eval, EvalCase

Eval(
    "MyProject",
    data=lambda: [EvalCase(input="a", expected="a")],
    task=lambda i, h: i,
    scores=[lambda input, output, expected, **_: 1.0 if output == expected else 0.0],
)
```

`eval_b.py`:
```python
from braintrust import Eval, EvalCase

Eval(
    "MyProject",
    data=lambda: [EvalCase(input="b", expected="b")],
    task=lambda i, h: i,
    scores=[lambda input, output, expected, **_: 1.0 if output == expected else 0.0],
)
```

Run:
```sh
braintrust eval eval_a.py eval_b.py --dev --dev-host 127.0.0.1 --dev-port 8300 --dev-org-name <my-org>
```

Startup log:
```
Loaded 2 evaluator(s): ['MyProject', 'MyProject']
```

`GET /list`:
```sh
$ curl -s http://127.0.0.1:8300/list \
    -H "Authorization: Bearer $BRAINTRUST_API_KEY" \
    -H "x-bt-org-name: <my-org>" | jq 'keys, length'
[ "MyProject" ]
1
```

Two evaluators loaded, only one reachable.

### Cause

`braintrust/devserver/server.py:308`:

```python
_all_evaluators = {evaluator.eval_name: evaluator for evaluator in evaluators}
```

The dict comprehension silently keeps the last entry per duplicate key. The startup log on the previous line prints the full input list *before* this collapse, which is why the loaded count looks correct.

Compounding factor: `Eval(name=...)` doubles as both the evaluator name and the project-name fallback (per the docstring at `framework.py:985`: "this corresponds to a project name in Braintrust"). The natural way to put multiple evaluators under one Braintrust project is to pass `Eval("SHARED_PROJECT_NAME", ...)` in each file — which produces this collision under `--dev`.

### Expected behavior

One of:
1. **Raise** at server startup naming the conflicting `eval_name` and the entrypoints that produced it.
2. **Warn** (logger or `warnings.warn`) on the collapse.
3. **Key** `_all_evaluators` by something more discriminating (e.g., file path + eval name).

Option 1 is the most defensible — silent data loss between the loader and the registered-evaluator dict shouldn't be possible. Option 3 is the most permissive but requires deciding how to disambiguate at the `/list` and `/eval` API.

### Workaround (verified)

Each entrypoint passes a distinct `name` AND an explicit `project_id` to keep multiple evaluators under one Braintrust project:

```python
from braintrust import Eval, EvalCase, init_dataset

_DATASET = init_dataset(project="MyProject", name="SomeDataset")

Eval(
    name="MyProject — variant_a",
    project_id=_DATASET.project_id,
    data=lambda: [EvalCase(input="a", expected="a")],
    task=lambda i, h: i,
    scores=[lambda input, output, expected, **_: 1.0 if output == expected else 0.0],
)
```

Verified empirically:
- `GET /list` returns two distinct entries (one per entrypoint with distinct `name`).
- Experiments from both evaluators land in the existing `MyProject` Braintrust project (URL pattern `…/p/MyProject/experiments/…`), confirming `project_id` overrides the `name`-based project resolution at write time (per `framework.py:767, 1792-1793`).

This is the supported pattern, but it's not discoverable: the API design (`name` doubles as the project-name fallback, no warning on dev-server collapse) leads users into the broken default.

### Fix sketch

In `braintrust/devserver/server.py`, replace the dict comprehension with a guarded loop:

```python
_all_evaluators: dict[str, Evaluator[Any, Any, Any]] = {}
for evaluator in evaluators:
    if evaluator.eval_name in _all_evaluators:
        raise ValueError(
            f"Multiple evaluators registered with name {evaluator.eval_name!r}. "
            f"Each --dev evaluator must have a distinct name. "
            f"Pass a unique `name=` to Eval(...) and `project_id=` to keep them in the same project."
        )
    _all_evaluators[evaluator.eval_name] = evaluator
```

Optionally, `framework.py`'s `Eval` could emit a deprecation note when `project_id` is omitted and `name` matches an existing Braintrust project — but that's a separate API discussion.

### Environment

- `braintrust` 0.17.0
- Python 3.14
- macOS

RitwijParmar:
## Summary

Fixes #366 by making duplicate `eval_name` registrations in the dev server addressable instead of silently collapsing to the last evaluator.

What changed:
- build the dev-server evaluator registry by stable evaluator id rather than raw `eval_name` only
- keep existing behavior for unique names: `/list` keys and `/eval { name }` continue to work
- when duplicate names exist, `/list` exposes disambiguated keys like `shared-project#1` / `shared-project#2` plus `id` and original `name` metadata
- `/eval` accepts the new `id` field, also accepts the disambiguated `/list` key as `name`, and returns a 409 with candidates when the original name is ambiguous

This avoids the startup-time raise/warn stopgap and keeps both evaluators reachable.

## Tests

- `PYTHONPATH=py/src .venv/bin/python -m pytest py/src/braintrust/devserver/test_server_integration.py::test_devserver_keeps_duplicate_eval_names_addressable py/src/braintrust/devserver/test_server_integration.py::test_eval_falls_back_to_evaluator_project_id_when_request_omits_or_empty_it py/src/braintrust/devserver/test_server_integration.py::test_eval_request_project_id_overrides_evaluator -q`
- `PYTHONPATH=py/src .venv/bin/python -m compileall -q py/src/braintrust/devserver/server.py py/src/braintrust/devserver/schemas.py py/src/braintrust/devserver/test_server_integration.py`
- `git diff --check`

### DIFF
--- a/py/src/braintrust/devserver/schemas.py
+++ b/py/src/braintrust/devserver/schemas.py
@@ -35,6 +35,7 @@ class ParsedEvalBody(TypedDict, total=False):
     """Type for parsed eval request body."""
 
     name: str  # Required
+    id: str
     parameters: dict[str, Any]
     data: Any
     scores: list[ParsedFunctionId]
@@ -201,6 +202,11 @@ def parse_eval_body(request_data: str | bytes | dict) -> ParsedEvalBody:
 
     # Build the parsed body
     parsed: ParsedEvalBody = {"name": name}
+
+    if "id" in data:
+        if not isinstance(data["id"], str):
+            raise ValidationError(f"id must be a string, got {type(data['id']).__name__}")
+        parsed["id"] = data["id"]
 
     # Optional fields with validation
     if "parameters" in data:
--- a/py/src/braintrust/devserver/server.py
+++ b/py/src/braintrust/devserver/server.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import sys
 import textwrap
+from collections import Counter, defaultdict
 from typing import Any
 
 
@@ -53,6 +54,7 @@ from .schemas import ValidationError, parse_eval_body
 
 
 _all_evaluators: dict[str, Evaluator[Any, Any, Any]] = {}
+_evaluator_ids_by_name: dict[str, list[str]] = {}
 
 
 class _ParameterOverrideHooks:
@@ -120,8 +122,10 @@ async def list_evaluators(request: Request) -> JSONResponse:
         return JSONResponse({"error": "Unauthorized"}, status_code=401)
 
     evaluator_list = {}
-    for name, evaluator in _all_evaluators.items():
-        evaluator_list[name] = {
+    for evaluator_id, evaluator in _all_evaluators.items():
+        ids_for_name = _evaluator_ids_by_name[evaluator.eval_name]
+        list_key = evaluator.eval_name if len(ids_for_name) == 1 else evaluator_id
+        evaluator_list[list_key] = {
             "parameters": (
                 serialize_remote_eval_parameters_container(evaluator.parameters) if evaluator.parameters else None
             ),
@@ -130,6 +134,9 @@ async def list_evaluators(request: Request) -> JSONResponse:
                 {"name": _classifier_name(classifier, i)} for i, classifier in enumerate(evaluator.classifiers or [])
             ],
         }
+        if len(ids_for_name) > 1:
+            evaluator_list[list_key]["id"] = evaluator_id
+            evaluator_list[list_key]["name"] = evaluator.eval_name
 
     return JSONResponse(evaluator_list)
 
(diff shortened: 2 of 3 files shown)


AbhiPrasad:
[automated] closing for repo clean up. Please re-open if this PR still applies!

PR #496, opened 2026-06-04
feat: otel compatible distributed tracing

realark:
- use otel-friendly IDs by default in the SDK
- add inject/extract methods for distributed tracing instead of parent slug methods
    - implements: https://www.w3.org/TR/trace-context/

OLD WAY: https://www.braintrust.dev/docs/instrument/advanced-tracing#trace-distributed-systems
```python
from braintrust import current_span, init_logger, start_span, traced

logger = init_logger(project="my-project")

# Client: Export the span
@traced
def process_request(request):
    return requests.post(
        "/api/process",
        json=request,
        headers={"X-Trace-ID": current_span().export()},
    )

# Server: Resume the trace
def handle_request(req):
    trace_id = req.headers.get("X-Trace-ID")

    with start_span(parent=trace_id) as span:
        result = process_data(req.body)
        span.log(input=req.body, output=result)
        return result
```

NOTE: the "old way" is still supported and backwards-compatible. Native SDKs can continue to link to each other in the manner, even if one is upgraded and the other is not. However if users with to link to otel sdks, they will have to change their code to use the new apis.

NEW WAY:
```python
from braintrust import current_span, init_logger, start_span, traced

logger = init_logger(project="my-project")

# Client: Export the span
@traced
def process_request(request):
    return requests.post(
        "/api/process",
        json=request,
        headers=current_span().inject({}), # <--- new inject api
    )

# Server: Resume the trace
def handle_request(req):
    with start_span(parent=extract_trace_context(req.headers) as span: # <--- new extract api
        result = process_data(req.body)
        span.log(input=req.body, output=result)
        return result
```

# Other notes

- confusing naming: this has the unfortunate effect of the `root_span_id` field essentially becoming the trace id rather than the id of the root span (i.e. `root_span_id` != `rootSpan.id`). This sucks, but it's already the case for otel sdks and py otel compat mode. So at least we're consistent.
- legacy id flag: the sdk currently maintains two paths for id generation. it defaults to UUIDs and does hex IDs when we're in in otel compat mode. This changes the default to use hex IDs and has a flag to revert to the old behavior. We could just rip out the old uuid logic entirely, but this approach seems like the least risky
- to support passing through tracestate and flags, an additional field was added to SpanImpl (propagated_state)

# Examples

- https://www.braintrust.dev/app/Braintrust%20SDKs/p/distributed-tracing-test/logs?range=%221h%22&r=021fb6471e30c686a667ffdfbba15ee6&s=67123a9c35899f93
- https://www.braintrust.dev/app/Braintrust%20SDKs/p/distributed-tracing-test/logs?range=%221h%22&r=5be69e7ec10f257b8671417162bcf075&s=050480159793924b

### DIFF
--- a/py/src/braintrust/propagation.py
+++ b/py/src/braintrust/propagation.py
@@ -1,0 +1,334 @@
+"""Native W3C Trace Context propagation for Braintrust.
+
+This module implements the propagation wire format described in the Braintrust
+distributed-tracing spec, using pure Python with no dependency on
+``opentelemetry``. It parses and serializes the W3C ``traceparent`` and
+``baggage`` headers and the Braintrust ``braintrust.parent`` baggage entry.
+
+Trace identity (trace id + parent span id) is carried in ``traceparent``; the
+Braintrust container the trace belongs to (project/experiment) is carried in
+``baggage`` under the ``braintrust.parent`` key.
+"""
+
+import logging
+import re
+from typing import NamedTuple
+from urllib.parse import quote, unquote
+
+
+__all__ = [
+    "TRACEPARENT_HEADER",
+    "TRACESTATE_HEADER",
+    "BAGGAGE_HEADER",
+    "BRAINTRUST_PARENT_KEY",
+    "DEFAULT_TRACE_FLAGS",
+    "ParsedTraceparent",
+    "PropagatedState",
+    "parse_traceparent",
+    "format_traceparent",
+    "parse_baggage",
+    "merge_baggage",
+    "get_header",
+]
+
+log = logging.getLogger(__name__)
+
+TRACEPARENT_HEADER = "traceparent"
+TRACESTATE_HEADER = "tracestate"
+BAGGAGE_HEADER = "baggage"
+BRAINTRUST_PARENT_KEY = "braintrust.parent"
+
+# Trace-flags byte we emit for traces we originate: sampled (low bit set).
+DEFAULT_TRACE_FLAGS = "01"
+
+
+class ParsedTraceparent(NamedTuple):
+    """Parsed W3C ``traceparent`` fields.
+
+    Tuple-compatible: unpacks as ``(trace_id, span_id, trace_flags)``.
+    ``trace_flags`` is the raw 2-hex trace-flags byte (e.g. ``"01"`` sampled,
+    ``"00"`` not sampled), kept raw so any future flag bits survive a
+    parse -> format round trip without per-bit handling.
+    """
+
+    trace_id: str
+    span_id: str
+    trace_flags: str
+
+
+class PropagatedState(NamedTuple):
+    """Inbound W3C trace-context state that Braintrust forwards but never interprets.
+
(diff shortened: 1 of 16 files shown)


py/src/braintrust/propagation.py
```diff
@@ -0,0 +1,251 @@
...
+# Encoded set is kept minimal: the characters that would otherwise break the
+# baggage grammar (member/list delimiters and the property separator) plus
+# ``%`` so the encoding is self-describing.
+_PERCENT_ENCODE = {
+    ",": "%2C",
+    ";": "%3B",
+    "=": "%3D",
+    " ": "%20",
+    "%": "%25",
+}
```
AbhiPrasad: this might not be enough to be spec compliant. Can we make sure this works w/ the baggage spec appropriately?

py/src/braintrust/propagation.py
```diff
@@ -0,0 +1,251 @@
...
+    Only the ``braintrust.parent`` member is (re)serialized, by us, from the
+    ``braintrust_parent`` argument. Any pre-existing ``braintrust.parent``
+    member in ``existing`` is dropped in favor of the supplied value.
+
+    Returns the merged header value, or None if there is nothing to emit (so
+    callers omit the header rather than emit an empty one).
+    """
+    members = []
+    if existing and isinstance(existing, str):
+        capped = existing[:_MAX_BAGGAGE_LENGTH] if len(existing) > _MAX_BAGGAGE_LENGTH else existing
```
AbhiPrasad: `existing[:_MAX_BAGGAGE_LENGTH]` is not safe, we might cut off a member mid value. Let's try to optimize for keeping the values around instead of sending partial values.

AbhiPrasad (commented):
I'm happy with this public API. Mind writing a spec doc for this so we can also be clear about the behavior difference between py/js and the otel powered sdks?

py/src/braintrust/propagation.py
realark: agreed, I update the implementation to use urllib instead. also added some more round-trip unit test cases

py/src/braintrust/propagation.py
realark: yeah this is too aggressive. I updated the logic to back up to the last completed value when we blow the max size limits

realark:
@AbhiPrasad sweet! I've been working on this spec in tandem with this PR: https://github.com/braintrustdata/braintrust-spec/pull/27

I just updated it to add test cases for encoding and size limits

py/src/braintrust/propagation.py
realark: (BTW these limits come from the baggage spec if you're curious: https://www.w3.org/TR/baggage/#limits )

Qard (approved):

AbhiPrasad (approved):

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.26.0

PR #508, opened 2026-06-11
 Strip nulls from remote eval prompt parameter defaults

evanmkeith:
## Summary

  Fix Python remote eval prompt parameter defaults that include explicit `None` / JSON `null` values from `PromptData.as_dict()`.

  `PromptData.as_dict()` currently preserves optional prompt fields like `tools`, `name`, `function_call`, and `tool_calls` as `None`. When these values are used as a `type:
  "prompt"` parameter default, the remote eval manifest contains explicit `null`s. The playground schema expects these fields to be omitted when unset, so the remote eval can fail
  to appear or fail to parse.

  This strips `None` values recursively only when serializing prompt parameter defaults, leaving the generic `PromptData.as_dict()` behavior unchanged.

  ## Test Plan

  - Add/updated Python parameter serialization test covering a `PromptData` prompt default with unset optional fields.
  - Verify serialized remote eval parameter container omits `tools`, `name`, `function_call`, and `tool_calls` when unset.
  - Manually verify `bt eval <repro>.py --dev` lists the remote eval in the playground.

### DIFF
--- a/py/src/braintrust/parameters.py
+++ b/py/src/braintrust/parameters.py
@@ -215,14 +215,32 @@ def is_eval_parameter_schema(schema: Any) -> bool:
     return True
 
 
+def _strip_none_values(value: Any) -> Any:
+    """Recursively drop dict entries whose value is ``None``.
+
+    Prompt parameter defaults are serialized from Python prompt dataclasses,
+    which preserve explicit ``None`` values for optional fields (e.g. a message's
+    ``name``/``function_call``/``tool_calls`` or a chat block's ``tools``). The
+    JS/UI evaluator manifest schema treats those fields as optional-when-absent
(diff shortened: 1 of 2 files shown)


j13huang (approved):
Okay, after mucking around a little bit, this looks pretty good. Thanks Evan. I made some frontend changes to support the null-fields format as well https://github.com/braintrustdata/braintrust/pull/15672


(For posterity, I explored some other serialization changes for SerializableDataClass but it's probably overkill https://github.com/braintrustdata/braintrust-sdk-python/commit/60ae80789657446e8526bc56f9ea0820c0b7cb89)

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.25.0

PR #512, opened 2026-06-12
fix(experiments): pass base_experiment_id to summarize

### LINKED ISSUE #511: Eval summary can compare against wrong baseline
### Description
`Eval` stores `base_experiment_id` correctly on the experiment but the final summary does not pass it as the explicit comparison ID. As a result, summary comparison can fall back to project/default baseline resolution and show wrong diffs.

### Reproduction
1. Navigate to Experiments in the Braintrust web app.
2. Select an Experiment and click **Set as default baseline**.
3. Run the following script:
```
import os
import uuid

import requests
from braintrust import Eval, EvalCase

PROJECT = f"<project-name>"
API_KEY = os.environ["BRAINTRUST_API_KEY"]
API_URL = os.environ.get("BRAINTRUST_API_URL", "https://api.braintrust.dev")
HEADERS = {"Authorization": f"Bearer {API_KEY}"}


def exact_match(input, output, expected):
    return 1.0 if output == expected else 0.0


def run_exp(name, output, base_experiment_id=None):
    result = Eval(
        PROJECT,
        experiment_name=name,
        data=[EvalCase(input="case", expected="yes")],
        task=lambda input: output,
        scores=[exact_match],
        base_experiment_id=base_experiment_id,
    )

    print(f"\n{name}")
    print(f"  id: {result.summary.experiment_id}")
    print(f"  compared_to: {result.summary.comparison_experiment_name}")
    print(f"  scores: {result.summary.scores}")
    return result


# B is the explicit baseline.
exp_b = run_exp("exp-B-baseline-score-0", "no")

# A should compare to B.
exp_a = run_exp(
    "exp-A-compare-to-B-score-1",
    "yes",
    base_experiment_id=exp_b.summary.experiment_id,
)

# Confirm A persisted B as its base_exp_id.
resp = requests.get(
    f"{API_URL}/v1/experiment/{exp_a.summary.experiment_id}",
    headers=HEADERS,
    timeout=30,
)
resp.raise_for_status()
record = resp.json()

print("\nPersisted base:")
print("  expected:", exp_b.summary.experiment_id)
print("  actual:  ", record.get("base_exp_id"))

print("\nReturned summary comparison:")
print("  expected:", exp_b.summary.experiment_name)
print("  actual:  ", exp_a.summary.comparison_experiment_name)
```

#### Expected
exp_a should be compared to exp_b

```
Persisted base:
  expected: <exp_b id>
  actual:   <exp_b id>

Returned summary comparison:
  expected: exp-B-baseline-score-0
  actual:   exp-B-baseline-score-0
```

#### Observed
exp_a is compared to the default baseline experiment selected in Step 2

barrettpyke:
### Description
Eval stores base_experiment_id correctly on the experiment but the final summary does not pass it as the explicit comparison ID. As a result, summary comparison can fall back to project/default baseline resolution and show wrong diffs.

### Fix
Pass `evaluator.base_experiment_id` into `experiment.summarize(comparison_experiment_id=...)`, so score and metric diffs are computed against the explicit experiment baseline.

Also resolve the explicit comparison experiment name so the returned summary displays the correct “compared to” name. Previously, `comparison_experiment_id` was `None`, so `summarize()` called `POST /api/base_experiment/get_id`; that resolver can apply UI/default-baseline behavior, including letting a project default baseline override the experiment’s explicit `base_exp_id`.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -1351,6 +1351,14 @@ def _validate_classification_result(value: Any, classifier_name: str) -> Classif
     return classification
 
 
+def _get_persisted_base_experiment_id(experiment: Experiment) -> str | None:
+    try:
+        base_experiment_id = experiment.data.get("base_exp_id")
+    except Exception:
+        return None
+    return base_experiment_id if isinstance(base_experiment_id, str) and base_experiment_id else None
+
+
 async def run_evaluator(
     experiment: Experiment | None,
     evaluator: Evaluator[Input, Output, Expected],
@@ -1367,9 +1375,12 @@ async def run_evaluator(
     )
 
     if experiment:
+        comparison_experiment_id = evaluator.base_experiment_id
+        if comparison_experiment_id is None:
+            comparison_experiment_id = _get_persisted_base_experiment_id(experiment)
         summary = experiment.summarize(
             summarize_scores=evaluator.summarize_scores,
-            comparison_experiment_id=evaluator.base_experiment_id,
+            comparison_experiment_id=comparison_experiment_id,
         )
     else:
         summary = build_local_summary(evaluator, results)
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -1367,7 +1367,10 @@ async def run_evaluator(
     )
 
     if experiment:
-        summary = experiment.summarize(summarize_scores=evaluator.summarize_scores)
+        summary = experiment.summarize(
+            summarize_scores=evaluator.summarize_scores,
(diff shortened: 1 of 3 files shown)


AbhiPrasad (commented):
`base_experiment_name` also has the same bug here. We need to fix that too

AbhiPrasad (approved):
I think the JS SDK has the same issue. Could you follow up with that?

barrettpyke:
@AbhiPrasad will do!

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.25.0

PR #532, opened 2026-06-25
feat(evals): Forward/pick up `bt eval <...> --sample N` flag

lforst:
Builds on https://github.com/braintrustdata/bt/pull/247

When doing `init_dataset()` with the `--sample N` flag, it will sample the rows.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -1808,12 +1808,12 @@ def init_dataset(
     """
 
     state = state or _state
-    sample_rate = getattr(builtins, "__bt_eval_sample_rate", None)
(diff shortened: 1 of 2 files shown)


py/src/braintrust/logger.py
```diff
@@ -1807,6 +1808,12 @@ def init_dataset(
     """
 
     state = state or _state
+    sample_rate = getattr(builtins, "__bt_eval_sample_rate", None)
```
ankrgyl: wonder if we should generalize this to `__bt_eval_internal_btql` so the cli can send more arbtirary stuff?

AbhiPrasad (approved):

PR #537, opened 2026-06-29
feat(prompts): add async prompt loading

### LINKED ISSUE #28: Async prompt loading support
## Summary

Add native async support for prompt loading in the Python SDK via an \`aload_prompt()\` function, allowing users to load multiple prompts in parallel using \`asyncio.gather\` without blocking.

## Context

- **Use case:** Users running async applications (e.g., with \`asyncio.gather\`) find that \`load_prompt()\` blocks, causing sequential execution when loading multiple prompts. This adds latency in production scenarios where prompts need to be fetched at request time.
- **Current workaround:** Use \`run_in_executor\` to wrap the sync call:

  ```python
  loop = asyncio.get_event_loop()
  prompt = await loop.run_in_executor(
      None,
      lambda: braintrust.load_prompt(project=project, slug=slug)
  )
  ```

- **Component**: Python SDK

## Prior Work

A prototype was built on branch `async-load-prompt` with usage:

```python
await braintrust.aload_prompt(project="...", slug="...")
```

This was not merged at the time, but multiple customers have since requested async support.

## Requirements

**Must have:**

- `aload_prompt()` function with same parameters as `load_prompt()`
- Works with `asyncio.gather` for parallel prompt loading
- Consistent behavior with sync version (caching, fallback, etc.)

## Related

- Linear issue: https://linear.app/braintrustdata/issue/BRA-3922/python-sdk-async-prompt-loading-support
- Related PR: https://github.com/braintrustdata/braintrust-sdk/pull/771

AbhiPrasad:
resolves https://github.com/braintrustdata/braintrust-sdk-python/issues/28

Add an awaitable prompt-loading API that mirrors `load_prompt()` while resolving prompt metadata off the event loop:

```python
prompt = await braintrust.load_prompt_async(project="My Project", slug="my-prompt")
kwargs = prompt.build(name="Ada")
```

The API can be used with `asyncio.gather()` to load multiple prompts without blocking the event loop:

```python
prompt_a, prompt_b = await asyncio.gather(
    braintrust.load_prompt_async(project="My Project", slug="prompt-a"),
    braintrust.load_prompt_async(project="My Project", slug="prompt-b"),
)
```

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -1,3 +1,4 @@
+import asyncio
 import atexit
 import base64
 import builtins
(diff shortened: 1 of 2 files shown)


py/src/braintrust/logger.py
```diff
@@ -2040,6 +2041,37 @@ def compute_metadata():
...
+        version=version,
+        project_id=project_id,
+        id=id,
+        defaults=defaults,
+        no_trace=no_trace,
+        environment=environment,
+        app_url=app_url,
+        api_key=api_key,
+        org_name=org_name,
+    )
```
realark, line 2077: should we add some kind of assertion or unit test to ensure load prompt and its async counterpart take identical arguments? thinking of cases where someone adds a new arg type to `load_prompt` and forgets to update the async fn

realark (approved):

PR #560, opened 2026-07-09
Fix Temporal replay span parent propagation

thibaultsoubeste:
Fix Temporal replay parent propagation for workflows that schedule work after replay.

What was happening:
- Workflow spans are intentionally not emitted again during replay.
- The outbound workflow interceptor still needs the workflow span context when it injects Braintrust headers into later activity, local activity, and child workflow commands.
- After replay, for example after worker cache eviction, newly scheduled activity spans could attach outside the `temporal.workflow.<Workflow>` span.

Fix:
- Add a Temporal patch marker for the new replay-safe propagation path.
- Use stable per-run workflow span ids keyed by Temporal namespace, workflow type, workflow id, and run id.
- Rebuild the workflow span context during replay, and use the same rebuilt context during the original execution so the propagated Braintrust header shape stays consistent.
- Add a regression workflow that forces replay with `max_cached_workflows=0`, resumes by signal, and checks post-replay activities stay under the single workflow span both with and without a client parent.

### DIFF
--- a/py/src/braintrust/integrations/temporal/plugin.py
+++ b/py/src/braintrust/integrations/temporal/plugin.py
@@ -80,6 +80,8 @@ The integration will automatically:
 """
 
 import dataclasses
+import hashlib
+import uuid
 from collections.abc import Mapping
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


AbhiPrasad:
hey @thibaultsoubeste thank you for the PR! I will review this tomorrow and work on getting it merged + released asap afterwards.

AbhiPrasad (approved):

AbhiPrasad:

PR #565, opened 2026-07-09
fix: log full-turn token usage from model_usage in claude-agent-sdk wrapper

### LINKED ISSUE #564: Claude Agent SDK: _handle_result logs main-agent usage, under-counts turns with subagents
## Summary

The Claude Agent SDK integration logs a turn's token metrics from `ResultMessage.usage`, which is the **main-agent-only** cumulative. On any turn that spawns subagents (the `Task`/Agent tool), the subagent tokens are billed but are missing from the trace, so the reported token total is a large under-count.

## Where

`braintrust/integrations/claude_agent_sdk/tracing.py`, `ContextTracker._handle_result` — the only place token metrics are attached:

```python
if hasattr(message, "usage"):
    usage_metrics, usage_metadata = extract_anthropic_usage(message.usage)
    ctx = self._get_context(None)
    if ctx.llm_span and (usage_metrics or usage_metadata):
        ctx.llm_span.log(metrics=usage_metrics or None, ...)
```

`ResultMessage.usage` is the cumulative usage for the **main agent**. `ResultMessage.model_usage` is the documented per-model breakdown that is **per-agent** — it includes subagent calls. The wrapper never reads `model_usage`.

## Reproduction

Requires an Anthropic API key. The script prints the token total the wrapper *would log* (from `usage`) next to the authoritative `model_usage` total for the same turn.

```python
# pip install claude-agent-sdk braintrust
# export ANTHROPIC_API_KEY=...
import anyio
from claude_agent_sdk import query, ClaudeAgentOptions, ResultMessage
from braintrust.wrappers.claude_agent_sdk import setup_claude_agent_sdk

# Instrument the SDK exactly as a Braintrust user would.
setup_claude_agent_sdk()

PROMPT = (
    "Use the Task tool to launch 3 subagents in parallel. Each subagent should "
    "independently run several tool calls exploring a different topic, then report "
    "back. Wait for all three, then synthesize. Actually spawn them via the Task tool."
)

def usage_total(u: dict) -> int:
    return (u.get("input_tokens", 0) + u.get("output_tokens", 0)
            + u.get("cache_read_input_tokens", 0) + u.get("cache_creation_input_tokens", 0))

def model_usage_total(mu: dict) -> int:
    return sum(m.get("inputTokens", 0) + m.get("outputTokens", 0)
               + m.get("cacheReadInputTokens", 0) + m.get("cacheCreationInputTokens", 0)
               for m in mu.values())

async def main():
    result = None
    async for msg in query(
        prompt=PROMPT,
        options=ClaudeAgentOptions(permission_mode="bypassPermissions"),
    ):
        if isinstance(msg, ResultMessage):
            result = msg

    logged = usage_total(result.usage)              # what _handle_result logs today
    authoritative = model_usage_total(result.model_usage)  # per-agent, incl. subagents
    print(f"wrapper-logged (usage):  {logged:,}")
    print(f"model_usage (per-agent): {authoritative:,}")
    print(f"wrapper captures:        {logged / authoritative:.1%}")

anyio.run(main)
```

**Observed** (one run, 3 subagents): `usage` = 126,113 vs `model_usage` = 873,617 → the wrapper captures **~14.4%** (two models: main Opus + Haiku subagents). In the logged trace the tokens land on a single LLM span; the subagent spans carry none. A run *without* subagents gives `usage == model_usage` (unaffected). Exact numbers vary per run; the direction does not.

## Cross-check that `model_usage` is the correct one

`model_usage × Anthropic list price` reconstructs `ResultMessage.total_cost_usd` (the SDK's own reported cost) exactly; the `usage`-based total reconstructs only a fraction. So the under-count is in what's logged, not in `model_usage`.

## Suggested fix

In `_handle_result`, derive the turn's token metrics from `model_usage` (aggregated across models) rather than `usage` — or attribute tokens onto the per-subagent spans. Anthropic's SDK documents `model_usage` as the authoritative per-agent source (anthropics/claude-agent-sdk-python#987).

PR: #565 — prefers the aggregated `model_usage` for the logged metrics, falls back to `usage` (metadata unchanged). Open to the per-subagent-span alternative if that's preferred.

## Related

- Same class of bug in the .NET SDK: braintrustdata/braintrust-sdk-dotnet#28, #45
- `model_usage` as authoritative per-agent source: anthropics/claude-agent-sdk-python#987

lsuarez-ekline:
## What

`ContextTracker._handle_result` logs the turn's token metrics from `ResultMessage.usage`, which is the cumulative usage for the **orchestrator agent only**. On a turn that spawns subagents (the `Task`/Agent tool), the subagent tokens are billed but missing from the trace, so the logged totals under-count — often heavily.

This prefers the token counts from `ResultMessage.model_usage` (the per-model breakdown, which is per-agent and includes subagent calls), summed across models, for the logged metrics. Metadata (service tier, inference geo) still comes from `usage`. When `model_usage` is missing or unusable it falls back to `usage`, so single-agent turns and older SDKs are unchanged.

Fixes #564.

## Details

- New helper `_aggregate_model_usage()` sums `model_usage` into an Anthropic-usage-shaped dict, then reuses `extract_anthropic_usage` for the metric mapping — so cache-read / cache-creation / completion all flow through the same path as before.
- Single-agent turns are unaffected (`usage == model_usage`).
- A model that only ever runs as a subagent has no dedicated LLM span; its tokens now land on the turn's LLM span via the aggregate, so the trace's token total matches the SDK's `model_usage`.

## Testing

- `nox -s "test_claude_agent_sdk(latest)"` — 50 passed (existing suite + 2 new unit tests for `_aggregate_model_usage`).
- `ruff format` / `ruff check` clean.

## Notes

Alternative considered: attribute tokens onto per-subagent spans instead of aggregating onto the turn's LLM span. Happy to take that direction if you'd prefer — this PR takes the minimal approach that makes the trace-level token total correct.


py/src/braintrust/integrations/claude_agent_sdk/tracing.py
```diff
@@ -729,6 +764,15 @@ def _handle_result(self, message: Any) -> None:
         self._active_key = None
         if hasattr(message, "usage"):
             usage_metrics, usage_metadata = extract_anthropic_usage(message.usage)
+            # ``usage`` covers only the orchestrator agent; prefer the token
+            # counts from ``model_usage`` (the per-model breakdown, which
+            # includes subagent calls) so the logged metrics reflect the whole
+            # turn. Metadata (service tier, etc.) still comes from ``usage``.
+            aggregated_usage = _aggregate_model_usage(getattr(message, "model_usage", None))
```
AbhiPrasad, line 771: unfortunately this will be incorrect because only one `metadata.model` is attached to this span. So if the subagent calls use different models then we calculate costs incorrectly.

AbhiPrasad (commented):
Thank you for the issue and PR! This will cause some bugs, so not gonna merge this in. Will open another PR to fix!

AbhiPrasad:
will be fixed by https://github.com/braintrustdata/braintrust-sdk-python/pull/670

PR #570, opened 2026-07-13
Add span origin provenance

Qard:
## Summary

Adds span origin provenance to the Python SDK direct logger and OpenTelemetry processor, including environment detection support.

## Validation

- python3 -m py_compile py/src/braintrust/env.py py/src/braintrust/logger.py py/src/braintrust/otel/__init__.py py/src/braintrust/span_origin.py
- aggregate git diff --check


py/src/braintrust/span_origin.py
```diff
@@ -0,0 +1,106 @@
...
+
+def _deployment_mode(value: str | None) -> SpanOriginEnvironment | None:
+    if not value:
+        return None
+    normalized = value.lower()
+    if normalized in ("production", "staging"):
+        return {"type": "server", "name": normalized}
+    if normalized in ("development", "local"):
+        return {"type": "local", "name": normalized}
+    return None
```
lforst, line 123: wondering whether we should always include name? the spec https://github.com/braintrustdata/braintrust-spec/pull/40/changes does not seem to say we should set name only if type is also defined

lforst (commented):
A small test would probably be good!

lforst (approved):

AbhiPrasad (approved):

py/src/braintrust/span_origin.py
lforst: @Qard sorry for the super late comment here, I missed this when I approved. I actually meant that we should probably just take PYTHON_ENV here verbatim for name as long as it's defined, and not make it dependent on it adhering to one of production staging development or local.

PR #595, opened 2026-07-17
fix: normalize list score dicts in remote evals

evrtt:
## Context

Valid multi-score remote scorer results from remote eval scorers are rejected. Remote eval scorers can return `list[Score]`, but that result crosses an HTTP/JSON boundary and the caller receives `list[dict]` rather than real `Score` objects. `await_or_run_scorer` normalizes a single score dict with `Score.from_dict()`, but is missing normalization for `list[dict]` responses. 
  
## Changes

  - Normalize dict entries inside scorer result lists with `Score.from_dict()`.
  - Keep existing validation for malformed score entries.
  - Add regression coverage for a scorer returning a list of score dicts.
  - Verify both per-result scores and summary scores are populated correctly.

  ## Testing

  * Added `test_run_evaluator_normalizes_list_of_dict_scores`
    - creates an evaluator with two data rows
    - uses a scorer that returns a list of two score dicts:
      - `{"name": "per_result", "score": 1.0}`
      - `"name": "summary_only", "score": 1.0}`
    - runs the evaluator locally with `run_evaluator(...)`
    - asserts each eval result contains both scores
    - asserts the final summary contains both score names with score 1.0

* Ran the focused core regression test
* Ran a local smoke test where a scorer returns multiple score dicts, confirming the eval completes and emits both expected scores.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -1463,17 +1463,17 @@ async def _run_evaluator_internal_impl(
             if isinstance(result, Iterable) and not isinstance(result, (str, bytes, Mapping)):
                 normalized_scores = []
                 for s in result:
+                    original_score = s
+                    score_error = None
                     if isinstance(s, dict):
                         try:
                             s = Score.from_dict(s)
                         except Exception as e:
-                            raise ValueError(
-                                f"When returning an array of scores, each score must be a valid Score object. Got: {s}"
-                            ) from e
-                    if not is_score(s):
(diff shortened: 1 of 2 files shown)


AbhiPrasad:
does this issue effect the other sdks?

evrtt:
@AbhiPrasad yeah Go and Ruby both similarly fail with a list of scores currently

AbhiPrasad (approved):

PR #613, opened 2026-07-22
fix: eval parameters omit null description

stretpjc:
resolves https://linear.app/braintrustdata/issue/SDK-92/remote-eval-params-with-no-description-serialize-description-null

## Summary

Remote eval dev mode (`braintrust eval <file> --dev`) hid **all** evals in the playground UI when any eval declared a parameter without a `description`.

All three branches of `serialize_eval_parameters` (prompt, model, data) emitted `"description": null` when unset. The playground's zod schema only accepts a string or an absent key, and it validates the whole `/list` response in one
`safeParse` — so one description-less parameter hid every eval on the endpoint.

Fix: include `description` only when it's not `None`, same as `default` already is. #508 fixed this pattern for prompt defaults; `description` was missed.

## Test 
With proposed change the remote eval is detected:
Screenshot 2026-07-22 at 2 13 25 PM

Same remote eval running in 0.30.1 is not detected:
image

### DIFF
--- a/py/src/braintrust/parameters.py
+++ b/py/src/braintrust/parameters.py
@@ -12,6 +12,7 @@ from typing_extensions import NotRequired, TypeGuard
 from .generated_types import PromptBlockDataNullish, PromptOptionsNullish
 from .prompt import PromptData
 from .serializable_data_class import SerializableDataClass
+from .util import clean_nones
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


stretpjc:
Forgot to mention the linear:
https://linear.app/braintrustdata/issue/SDK-92/remote-eval-params-with-no-description-serialize-description-null

AbhiPrasad (approved):

PR #624, opened 2026-07-29
fix(langchain): prevent async runs from inheriting ended spans

### LINKED ISSUE #621: Async LangChain runs nest inside each other instead of producing separate traces
**SDK version:** braintrust 0.30.1 (latest) · Python 3.12 · langchain-core 1.x

## What happens

When an application runs several LangChain/LangGraph invocations one after another using the async API, only the *first* one shows up as a trace. Every later invocation is logged as a child of the previous one.

So instead of this, which is what you'd expect:

```
run0
run1
run2
```

we get this:

```
run0
└── run1
    └── run2
```

The same code using the synchronous `.invoke()` API behaves correctly and produces three separate traces. Only the async `.ainvoke()` path is affected. A long-running worker keeps nesting forever.

## Minimal reproduction

No API key or network required.

```bash
pip install braintrust==0.30.1 langchain-core
python bt_repro.py
```

```python
"""Consecutive top-level async LangChain runs nest instead of being siblings."""
import asyncio

from braintrust import current_span, logger as bt_logger
from braintrust.integrations.langchain import BraintrustCallbackHandler
from braintrust.test_helpers import init_test_logger, simulate_login
from langchain_core.runnables import RunnableLambda

simulate_login()
chain = RunnableLambda(lambda x: x)


async def main():
    with bt_logger._internal_with_memory_background_logger() as bgl:
        init_test_logger("repro")
        handler = BraintrustCallbackHandler()

        for i in range(3):
            await chain.ainvoke(
                i, config={"callbacks": [handler], "run_name": f"run{i}"}
            )
            print(f"after run{i}, current span is still open: "
                  f"{type(current_span()).__name__}")

        spans = bgl.pop()
        roots = [s for s in spans if s.get("span_parents") is None]
        print(f"\n{len(spans)} spans, {len(roots)} roots (expected 3 roots)")
        for s in spans:
            print(f"  {s['span_attributes']['name']:6} id={s['span_id']} "
                  f"parents={s.get('span_parents')}")


asyncio.run(main())
```

**Actual output:**

```
after run0, current span is still open: SpanImpl
after run1, current span is still open: SpanImpl
after run2, current span is still open: SpanImpl

3 spans, 1 roots (expected 3 roots)
  run0   id=c9f2e7e35685a79d parents=None
  run1   id=e4237da31be9ddc6 parents=['c9f2e7e35685a79d']
  run2   id=87d1280aa4b77596 parents=['e4237da31be9ddc6']
```

**Change the three `await chain.ainvoke(...)` calls to `chain.invoke(...)` and
the same script prints:**

```
after run0, current span is still open: _NoopSpan
after run1, current span is still open: _NoopSpan
after run2, current span is still open: _NoopSpan

3 spans, 3 roots (expected 3 roots)
  run0   id=f50379a9aca7b218 parents=None
  run1   id=2e282f984cac5398 parents=None
  run2   id=c9e8edba7233a436 parents=None
```

The `current span is still open: SpanImpl` line shows that after an async run finishes, a span that has already ended is still registered as the current one.

## Workaround we're using

Running each top-level invocation in its own `asyncio.Task`. A task gets a copy of the context, so a leaked span dies with the task instead of bleeding into the next run — and because it's a copy rather than a reset, deliberately-opened parent spans are still inherited correctly.

```python
def isolated_trace_context(fn):
    @functools.wraps(fn)
    async def wrapper(*args, **kwargs):
        return await asyncio.create_task(fn(*args, **kwargs))
    return wrapper
```

AbhiPrasad:
fixes https://github.com/braintrustdata/braintrust-sdk-python/issues/621

### DIFF
--- a/py/src/braintrust/context.py
+++ b/py/src/braintrust/context.py
@@ -24,6 +24,12 @@ class ParentSpanIds:
     span_parents: list[str]
 
 
+@dataclass(frozen=True)
+class _SpanContextEntry:
+    span_object: Any
+    parent: "_SpanContextEntry | None"
+
+
 class ContextManager(ABC):
     """Abstract base class for managing span context in Braintrust.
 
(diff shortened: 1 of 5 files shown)


lforst (approved):

AbhiPrasad:
Fix for this was released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.31.0!

PR #640, opened 2026-07-30
fix: surface experiment-comparison2 fetch failures and retry transient 5xx

### LINKED ISSUE #639: Experiment.summarize() silently discards scores/metrics on any fetch failure instead of surfacing or retrying
## Summary

## Where

```python
try:
        "experiment-comparison2",
    )
except Exception as e:
    _logger.warning(
    )
    summary_items = {}
```

## Repro

```python
import braintrust

```

## Suggested fix

Either or both:

## Environment

- `braintrust` (PyPI) 0.29.0

praneethy91:
## Summary

Fixes #639.

`Experiment.summarize()` caught *any* exception from the `experiment-comparison2` fetch and silently fell back to an empty `{}` summary, making a transient fetch failure indistinguishable from an experiment that genuinely has no scores. Meanwhile, `RetryRequestExceptionsAdapter` only retried on connection-level exceptions (timeouts, connection resets), so a request that completed with a non-2xx status (e.g. a 502 from a proxy hiccup) never entered the retry path at all — it went straight to `raise_for_status()` and got swallowed by `summarize()`.

- `RetryRequestExceptionsAdapter.send()` now also retries when a completed response has a transient status code (429/500/502/503/504), using the same exponential backoff as the existing exception-based retry path. This should eliminate most one-off 5xx blips before they ever reach `summarize()`.
- `ExperimentSummary` gains a new `scores_fetch_error: str | None` field, populated whenever the `experiment-comparison2` fetch fails after retries are exhausted. `scores`/`metrics` still default to `{}` for backwards compatibility, but callers building automation (e.g. CI gates) on `summary.scores` can now check `scores_fetch_error` to tell "fetch failed" apart from "no scores." The printed `str(summary)` also surfaces a `WARNING` line when this happens.

## Test plan

- [x] `py/src/braintrust/test_http.py::TestRetryOnHttpErrorStatus` — new tests covering retry-on-502, giving up after `base_num_retries`, and *not* retrying non-retryable statuses (404), using a real local HTTP server (no mocks), consistent with existing adapter tests in this file.
- [x] `py/src/braintrust/test_framework.py::test_experiment_summarize_surfaces_scores_fetch_error` — new test asserting `summarize()` populates `scores_fetch_error` (and the printed summary contains a `WARNING`) when the comparison fetch raises.
- [x] `cd py && uv run nox -s test_core` — full core suite passes (614 passed, 63 skipped, 12 xfailed).
- [x] `cd py && uv run nox -s pylint` — clean.
- [x] `cd py && uv run pre-commit run --files ...` on changed files — clean.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -728,16 +728,23 @@ def set_http_adapter(adapter: HTTPAdapter) -> None:
         _state._api_conn._reset()
 
 
+#: HTTP status codes that indicate a transient failure worth retrying, rather than a
+#: client error that will fail identically on every attempt.
+RETRYABLE_STATUS_CODES = frozenset({429, 500, 502, 503, 504})
+
+
 class RetryRequestExceptionsAdapter(HTTPAdapter):
     """An HTTP adapter that automatically retries requests on connection exceptions.
 
     This adapter extends requests' HTTPAdapter to add retry logic for common network-related
-    exceptions including connection errors, timeouts, and other HTTP errors. It implements
-    an exponential backoff strategy between retries to avoid overwhelming servers during
-    intermittent connectivity issues.
+    exceptions including connection errors, timeouts, and other HTTP errors, as well as
+    responses that complete with a transient HTTP error status (see `RETRYABLE_STATUS_CODES`).
+    It implements an exponential backoff strategy between retries to avoid overwhelming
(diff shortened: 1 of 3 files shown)


AbhiPrasad:
will be fixed by https://github.com/braintrustdata/braintrust-sdk-python/pull/682, closing this PR as a result

PR #669, opened 2026-08-10
ref: replace GitPython with direct git commands

AbhiPrasad:
resolves https://linear.app/braintrustdata/issue/SDK-211/make-gitpython-an-optional-extra-in-braintrust-sdk

in order to validate this was legit, I (well mr codex) added extra test coverage in `py/src/braintrust/test_git_metadata_vcr.py`, and then did the gitpython migration accordingly.


viadezo1er (approved):
stamp

PR #670, opened 2026-08-10
fix(claude_agent_sdk): stop assigning query usage to the final call

### LINKED ISSUE #564: Claude Agent SDK: _handle_result logs main-agent usage, under-counts turns with subagents

AbhiPrasad:
resolves #564
resolves https://linear.app/braintrustdata/issue/SDK-52/claude-agent-sdk-handle-result-logs-main-agent-usage-under-counts

## AI Summary

`ResultMessage.usage` is cumulative across a Claude Agent SDK query, so attaching it to the final `anthropic.messages.create` span misattributes root and subagent tokens and produces incorrect prompt-cache costs.

Token accounting now follows the caller's existing `include_partial_messages` setting without mutating options, injecting hooks, or reading transcript files:

```
include_partial_messages=True
  message_start/message_delta -> exact per-request usage on each LLM span
  result                      -> query metadata only

include_partial_messages=False or omitted
  assistant -> no token metrics on individual LLM spans
  result    -> aggregate all-agent usage on the root "Claude Agent" task span
```

For aggregate task usage, the integration prefers `ResultMessage.model_usage`, which includes root agents and subagents across models, and falls back to `ResultMessage.usage` when model usage is unavailable. Allowlisted result usage metadata such as `service_tier` and `inference_geo` is preserved on the task span.

Caller-provided options and hooks remain unchanged, and yielded SDK messages are not modified. The behavior is covered through both the one-shot `query()` helper and `ClaudeSDKClient`, including direct `connect(prompt)` calls, multiple provider requests, bundled subagents, hooks, and concurrent/racing MCP tool handling.

Anthropic usage normalization now emits either TTL-specific cache-creation metrics or the legacy aggregate cache-creation metric, never both. Prompt totals use the same representation, and total token counts are emitted only when completion usage is known.

Transport cassettes and tests cover the latest Claude Agent SDK and the pinned 0.1.10 compatibility version.

### DIFF
--- a/py/src/braintrust/integrations/anthropic/_utils.py
+++ b/py/src/braintrust/integrations/anthropic/_utils.py
@@ -53,11 +53,18 @@ def _set_numeric_metric(metrics: dict[str, float], name: str, value: Any) -> Non
         metrics[name] = float(value)
 
 
-def extract_anthropic_usage(usage: Any) -> tuple[dict[str, float], dict[str, Any]]:
+def extract_anthropic_usage(
+    usage: Any,
+    *,
+    include_output: bool = True,
+    include_legacy_cache_creation: bool = True,
+) -> tuple[dict[str, float], dict[str, Any]]:
     """Extract normalized metrics and allowlisted metadata from Anthropic usage.
 
     Numeric usage fields are converted into Braintrust metrics. Allowlisted
     non-numeric fields are attached as span metadata with a ``usage_`` prefix.
+    Anthropic's per-TTL cache creation breakdown supersedes the legacy aggregate
+    metric, and totals are emitted only when completion usage is known.
     """
     usage = _try_to_dict(usage)
     if usage is None:
@@ -66,6 +73,10 @@ def extract_anthropic_usage(usage: Any) -> tuple[dict[str, float], dict[str, Any
     metrics: dict[str, float] = {}
     metadata: dict[str, Any] = {}
     for source_name, metric_name in _ANTHROPIC_USAGE_METRIC_FIELDS:
+        if metric_name == "completion_tokens" and not include_output:
+            continue
+        if metric_name == "prompt_cache_creation_tokens" and not include_legacy_cache_creation:
+            continue
         _set_numeric_metric(metrics, metric_name, usage.get(source_name))
 
     cache_creation = _try_to_dict(usage.get("cache_creation"))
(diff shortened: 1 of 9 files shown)


py/src/braintrust/integrations/anthropic/_utils.py
```diff
@@ -77,22 +88,36 @@ def extract_anthropic_usage(usage: Any) -> tuple[dict[str, float], dict[str, Any
                 metrics[metric_name] = float(value)
                 cache_creation_breakdown.append(float(value))
 
+    if cache_creation_breakdown:
+        metrics.pop("prompt_cache_creation_tokens", None)
```
realark: is it worth some kind of guard to make sure that the sum of all the specific metrics we collect equals anthropic's total count? thinking of cases where they add another field that's not in `_ANTHROPIC_CACHE_CREATION_METRIC_FIELDS`

py/src/braintrust/integrations/anthropic/_utils.py
```diff
@@ -53,11 +53,18 @@ def _set_numeric_metric(metrics: dict[str, float], name: str, value: Any) -> Non
         metrics[name] = float(value)
 
 
-def extract_anthropic_usage(usage: Any) -> tuple[dict[str, float], dict[str, Any]]:
+def extract_anthropic_usage(
+    usage: Any,
+    *,
+    include_output: bool = True,
+    include_legacy_cache_creation: bool = True,
```
realark: do any callers override this flag?

lforst (approved):

Issue #673, opened 2026-08-10
Python API client architecture and migration plan


AbhiPrasad:
Status: Implementation in progress

## Implementation progress

- [X] PR 1: Extract the existing transport without changing behavior — [#641](<https://github.com/braintrustdata/braintrust-sdk-python/pull/641>), merged in [`625a5e77bcd0192bae4aa8e30c86e753d34187d2`](<https://github.com/braintrustdata/braintrust-sdk-python/commit/625a5e77bcd0192bae4aa8e30c86e753d34187d2>) (`ref(api): establish transport boundary for API client`).
- [X] PR 2: Add structured errors and the policy-aware request engine — merged in [`0cbf393e8992c756e3d10e16be4be9eb659c9abe`](<https://github.com/braintrustdata/braintrust-sdk-python/commit/0cbf393e8992c756e3d10e16be4be9eb659c9abe>) (`feat(api): add policy-aware HTTP transport`, [#653](<https://github.com/braintrustdata/braintrust-sdk-python/pull/653>)).
- [X] PR 3: Add client bootstrap and service facade — merged in [`aa1f21ea0ecbb18f7f49c0cff6a5ab7b9c9b7392`](<https://github.com/braintrustdata/braintrust-sdk-python/commit/aa1f21ea0ecbb18f7f49c0cff6a5ab7b9c9b7392>) (`feat(api): add client bootstrap and service facade`, [#655](<https://github.com/braintrustdata/braintrust-sdk-python/pull/655>)).
- [ ] PRs 4–12 and public API stabilization — not started.

## Context

The Python SDK does not currently have an API module comparable to the Go SDK's `api/` package. HTTP transport, authentication, endpoint calls, response parsing, retries, and product-level fallback behavior are spread across `logger.py`, `functions/invoke.py`, `framework2.py`, `sandbox.py`, the devserver, and CLI modules.

This became especially visible in [issue #639](<https://github.com/braintrustdata/braintrust-sdk-python/issues/639>): `Experiment.summarize()` catches any failure from `experiment-comparison2` and returns empty score and metric maps. That makes a failed read look like a successful experiment with no scores.

An audit of the Python SDK's Braintrust-directed HTTP calls found a broader transport problem:

* Before login, connections have no default retries.
[...]

PR #682, opened 2026-08-12
feat(api): migrate experiments to the API client

### LINKED ISSUE #639: Experiment.summarize() silently discards scores/metrics on any fetch failure instead of surfacing or retrying

AbhiPrasad:
ref https://github.com/braintrustdata/braintrust-sdk-python/issues/673
supercedes https://github.com/braintrustdata/braintrust-sdk-python/pull/640

Migrate experiment lookup, base resolution, and comparison reads to the policy-aware API workflow with safe-read retries and typed responses.

Resolve https://github.com/braintrustdata/braintrust-sdk-python/issues/639 by retrying transient comparison failures and raising typed API errors after retries are exhausted instead of presenting failures as empty scores. Represent successful and intentionally skipped summaries explicitly.

### DIFF
--- a/py/src/braintrust/api/__init__.py
+++ b/py/src/braintrust/api/__init__.py
@@ -14,7 +14,14 @@ from .errors import (
     BraintrustTransportError,
     BraintrustTransportRetryExhaustedError,
 )
-from .experiments import ExperimentsAPI
+from .experiments import (
+    BaseExperiment,
+    ExperimentComparison,
+    ExperimentMetric,
+    ExperimentRecord,
+    ExperimentsAPI,
+    ExperimentScore,
+)
 from .functions import FunctionsAPI
 from .policies import RetryMode, RetryPolicy
 from .projects import ProjectsAPI
@@ -25,6 +32,7 @@ from .queries import QueriesAPI
 __all__ = [
     "AttachmentsAPI",
     "AuthAPI",
+    "BaseExperiment",
     "BraintrustAPIError",
     "BraintrustClient",
     "BraintrustHTTPError",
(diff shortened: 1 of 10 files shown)


AbhiPrasad:
Closing in favor of the generated API client — tracked in #683.

**Why:** this PR returns frozen dataclasses (`ExperimentRecord`, `ExperimentComparison`, …) with
`from_dict` validation and a `raw` verbatim-body passthrough. #683 generates TypedDicts from
[`braintrust-openapi`](https://github.com/braintrustdata/braintrust-openapi). Merging this would put
two incompatible return conventions in `braintrust.api`, so the convention question is settled up
front instead of after the fact: **generated TypedDicts**, with additive-field tolerance coming from
non-closed TypedDicts rather than a `raw` field. Recorded as open decision 7 in #683.

Nothing here is wasted — the branch stays readable and is the reference implementation for the redo.
#683 records these as requirements it must satisfy:

- `get_base()` returning `None` on HTTP 400 (a fresh project's first eval always takes this path, so
  generated bindings must not flatten it into a generic 4xx error);
- `get()` / `get_base()` / `compare()` classified `RetryMode.SAFE_READ` — the concrete case of the
  call-site audit's POST-based-logical-read class;
- additive backend fields surviving round-trips;
- `SummarySuccess` / `SummarySkipped` in `logger.py`;
- the coverage shape — real-backend VCR cassettes plus a local `http.server` test asserting exact
  request targets and additive-field preservation, plus type tests — as the template for every
  subsequent resource migration.

**Note on #639:** this PR was its only fix in flight, and that bug is user-facing — a failed
comparison read currently looks identical to a successful experiment with no scores. #683 tracks two
ways to unpark it: re-land only the product-level half now (the strict-by-default semantics and the
discriminated `SummarySuccess`/`SummarySkipped`/`SummaryFailed` result in `logger.py`, which are
eval-summary types rather than REST wire types and so do not touch the convention decision), or wait
for generated Experiments bindings in step 3.

Issue #683, opened 2026-08-12
OpenAPI-generated Python API client plan


AbhiPrasad:
Tracking issue for generating the public Braintrust REST API client from  <br>[`braintrust-openapi`](<https://github.com/braintrustdata/braintrust-openapi>) and exposing it through  <br>`braintrust.api` without changing existing authentication, routing, retries, or high-level SDK behavior.

This supersedes braintrustdata/braintrust-sdk-python#673. Keep braintrustdata/braintrust-sdk-python#673 as the detailed runtime/call-site audit and braintrustdata/braintrust-sdk-python#682 as the reference for  <br>the Experiments behavior and tests. This issue is the implementation source of truth.

## Implementation checklist

- [x] Foundation: transport (#641), policy-aware errors/retries (#653), and client/service facade (#655)
- [x] Generator foundation: measurement spike, pinned spec/models/drift CI (#690), and cross-resource partitioning
- [x] `Projects` bindings: generate all five operations (#697)
- [x] `Projects` SDK migration: migrate registration and project-lookup call sites (#705; Dataset project resolution migrated in #703)
- [x] `Experiments` bindings and summary migration: generate all ten operations and migrate `Experiment.summarize()` (#701)
- [x] `Experiments` SDK migration: migrate registration, lookup, and ordinary fetching ([509610f8](https://github.com/braintrustdata/braintrust-sdk-python/commit/509610f8eb256c694134ce0e2d27aa6d59d70066))
- [x] `Datasets`: generate all ten operations and migrate registration, ordinary fetching, summaries, and devserver ID lookup (#703)
- [x] `Prompts`: generate all six operations and migrate `load_prompt()` retrieval (#750)
- [x] `Functions`: generate all six metadata CRUD operations and migrate `load_parameters()` retrieval (#758); keep invocation specialized
- [ ] Generate all remaining reviewed REST tags; keep intentionally unsupported tags documented and exclude Proxy
- [ ] Document the public REST resources, types, and stability policy
- [ ] Automate pinned-spec update PRs after 3–4 manual updates

Next: **generate the remaining reviewed REST tags; Projects, Experiments, Datasets, Prompts, and Functions are complete.**

## Goal

[...]

PR #701, opened 2026-08-18
feat(api): generate Experiments REST bindings

### LINKED ISSUE #639: Experiment.summarize() silently discards scores/metrics on any fetch failure instead of surfacing or retrying

AbhiPrasad:
Implements step 3 of #683 and fixes #639.

### AI Summary

This adds the second public generated REST resource:

- `BraintrustOpenApiClient.experiments`: all 10 operations selected by the Experiments OpenAPI tag, including `get_experiment_id_summarize`
- `braintrust.api.types`: public experiment request and response types


openapi/config.json
```diff
@@ -30,7 +30,11 @@
   "endpoint_generator": {
     "schema_version": 1,
     "generated_tags": [
-      "Projects"
+      "Projects",
+      "Experiments"
```
lforst, line 34: if I understand correctly, I wonder if we will ever leak unwanted endpoints into the generated client

lforst (approved):

openapi/config.json
lforst: or accidentally remove endpoints because someone removed the tag and break them

AbhiPrasad: I did the allowlist this way so that we very explicitly add only the endpoints we want to the generated client.

Removing tags should also break tests so I'm not too worried about that.

PR #762, opened 2026-09-11
feat(openai): Add Agents API instrumentation

AbhiPrasad:
resolves https://linear.app/braintrustdata/issue/SDK-354/add-openai-agents-api-integration-to-python-sdk

Instruments `openai.agents.sessions.create`

image

Right now this has some limitations

1. openai doesn't give us `usage` details consistently on it's session events, so we can't reliably attach token counts to spans 😢
2. we don't get detailed spans about subagents

### DIFF
--- a/py/pyproject.toml
+++ b/py/pyproject.toml
@@ -377,7 +377,7 @@ latest = "ai[openai]==0.5.2"
 "0.4.0" = "ai[openai]==0.4.0"
 
 [tool.braintrust.matrix.openai]
-latest = "openai==3.8.0"
+latest = "openai==3.13.0"
 "1.92.0" = "openai==1.92.0"
 "1.77.0" = "openai==1.77.0"
 "1.71.0" = "openai==1.71.0"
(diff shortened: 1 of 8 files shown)


py/pyproject.toml
```diff
@@ -377,7 +377,7 @@ latest = "ai[openai]==0.5.2"
 "0.4.0" = "ai[openai]==0.4.0"
 
 [tool.braintrust.matrix.openai]
-latest = "openai==3.8.0"
+latest = "openai==3.13.0"
```
realark, line 380: I assume the goal here is to just keep bumping up `latest` and if/when we hit some kind of major compat issue only then do we add a new entry to the matrix

realark (approved):

py/pyproject.toml
AbhiPrasad: yup exactly. Every week we have a weekly job to update it. We still prefer fixed version numbers over just saying whatever is latest to prevent supply chain attacks.

PR #782, opened 2026-09-18
perf: cache SDK version for span origin context

nathan-mialabs:
cache installed Braintrust SDK version for the lifetime of the process
add regression test proving repeated span-origin merges resolve package metadata only once

why:

`merge_span_origin_context()` runs for every span, but currently resolves the installed Braintrust version through `importlib.metadata` each time. That triggers synchronous `os.stat()` calls. We observed this with AIOCOP, an asyncio event-loop monitoring tool: these synchronous calls appeared in stalls lasting 2.45–3.18 seconds under load.

Since the installed version cannot change during a process's lifetime, caching it removes repeated blocking I/O while preserving behavior.

Example stack trace captured in SigNoz:

```text
importlib_metadata/__init__.py:838:mtime
<- importlib_metadata/__init__.py:833:search
<- importlib_metadata/__init__.py:974:<genexpr>
<- importlib.metadata:397:from_name
<- importlib.metadata:862:distribution
<- importlib.metadata:889:version
<- braintrust/span_origin.py:103:_sdk_version
<- braintrust/span_origin.py:92:merge_span_origin_context
<- braintrust/logger.py:4619:__init__
<- braintrust/logger.py:4761:start_span
<- braintrust/logger.py:2981:start_span
```

timing:

Local timing on macOS with Python 3.12 using `python -m timeit`:

- uncached `importlib.metadata.version("braintrust")`: 113 µs per call
- cached `_sdk_version()`: 10.3 ns per call
- approximately 11,000x lower hot-path overhead after the first lookup


AbhiPrasad (approved):

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-python/releases/tag/py-sdk-v0.41.1

